PR #22, opened 2020-10-23
Add country as a council attribute

### LINKED ISSUE #21: Add countries

crowbot:
Fixes #21


zarino (approved):
Not sure there’s much I’m qualified to review here, but the migration applies cleanly with `script/migrate` and the server runs, so I assume it’s all good.

PR #39, opened 2020-11-03
Solr fulltext search

### LINKED ISSUE #28: Add full text search

crowbot:
Adds fulltext search using [Solr 6.6](https://lucene.apache.org/solr/guide/6_6/) through [Haystack 3](https://django-haystack.readthedocs.io/en/master/)

I've added some example parameters passed to Solr for the snippets and highlighting in the text search results in `HighlightedSearchForm` in `caps/forms`. Through these parameters, we can change (I think) anything listed in 
[https://lucene.apache.org/solr/guide/6_6/highlighting.html#Highlighting-Usage](https://lucene.apache.org/solr/guide/6_6/highlighting.html#Highlighting-Usage)

Not sure about adding integration tests here - I feel a bit guilty about not adding any, but there's so little code around the search that they'd be testing the code, so much as testing that solr is present and the data is indexed. Feels like trying to write tests for that might just add complexity without adding much value.

Closes #28

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -32,8 +32,8 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install packages
       run: |
+        xargs sudo apt-get update
         xargs sudo apt-get install -qq -y < conf/packages
-        dpkg -l '*poppler*'
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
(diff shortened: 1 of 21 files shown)


sagepe (approved):
I think this looks good. I've gone through the haystack/solr integration while looking through the docs, and it all seems simple and straightforward.

I wonder how often we are likely to need to regenerate the solr XML files - I'm assuming that these won't actually change very often.

PR #43, opened 2020-11-03
Set up styles and templates

zarino:
I appreciate the file diff is a bit intimidating here, because of the new `/vendor` directory. Here‘s a summary of the changes:

* Moves HTML templates into the `/caps` app.
* Adds a `base.html` template that all the other HTML templates extend.
* Adds bootstrap sass and js, jQuery js, and html5shiv js in a new `/vendor` directory.
* Adds a `/caps/static/scss/main.scss` file that imports a selection of bootstrap modules, and overrides an example variable, just to check everything’s working.
* Makes sure that `django.contrib.staticfiles` and `django-pipeline` are configured to play nicely together – staticfiles collects all the scss/js files into `/static`, and then django-pipeline compiles the scss files from there.
* Uses `django-bootstrap4` to add bootstrap classes/structure to django forms.

It felt sensible to move the templates into the `/caps` app sooner rather than later. But this might cause a temporary headache when merging the pull requests that are already in flight. I did a `git mv` in a commit of its own, so _hopefully_ Git will work out what’s happened. But @crowbot let me know if you think it’ll cause problems. Maybe we could merge all the other PRs, then rebase this on top?


crowbot (approved):
Looks good to me, and works locally. The styling on the filters when you select one and submit is a bit odd - they're all getting the 'correctly filled in' green border and check mark.

zarino:
> The styling on the filters when you select one and submit is a bit odd - they're all getting the 'correctly filled in' green border and check mark.

Whoa yeah, that is weird. I wonder whether we can tell `django-filter` / `django-bootstrap4` not to "validate" the form…

zarino:
@LouiseMySociety I‘ve pushed a commit that prevents `django-bootstrap4` from generating the `is-valid` styles, across the whole site.

PR #47, opened 2020-11-04
Handle a plan that can't be downloaded without erroring

### LINKED ISSUE #25: Better handle case where plan document URL errors
This can happen for a newly added invalid URL, or an existing one where the document is no longer available, or a transient error. At the moment, an error message is [shown in the terminal](https://github.com/mysociety/caps/blob/master/caps/management/commands/preprocess.py#L82) but later code expects the plan document to be stored locally, and terminates with an exception, eg. https://github.com/mysociety/caps/issues/24

Initial thought is to remove the URL in the CSV - but is that the right thing to do when a URL for an already downloaded plan starts to error?

crowbot:
Just show an error message and continue. Fixes #25

### DIFF
--- a/caps/management/commands/preprocess.py
+++ b/caps/management/commands/preprocess.py
@@ -9,6 +9,7 @@ from urllib.parse import urlparse
 import urllib3
 
 import pandas as pd
+import numpy
 
 from django.core.management.base import BaseCommand, CommandError
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


sagepe (approved):
LGTM. I've cherry-picked this and tested and it completes with a bad URL successfully.

PR #48, opened 2020-11-04
Infra improvements

sagepe:
This does the following:

* Updates the Vagrant box to a minimal-install of Debian Buster
* Cleans up the various package files a bit
* Removes some superfluous arguments and commands from the Vagrantfile
* Removes the virtualenv

This last may seem a little odd, but we're planning on running this in containers in production which will take the place of the virtualenv in terms of environment isolation; this is a step towards that.


crowbot (approved):
👍 Looks good to me!

PR #49, opened 2020-11-04
Add a faster version of the slow bits of script/update

### LINKED ISSUE #36: Add PDF extension to PDF files

### LINKED ISSUE #42: Don't make requests for downloaded PDFs in script/update
Currently HEAD requests are made if files are already downloaded locally in order to set the filetype and charset of files. We could get the filetype info by setting an extension on the files downloaded and using that.

crowbot:
This allows cached data to be used on a standard run of `script/update` (if any data has previously been loaded). A run with the flag `script/update --all` will produce a comprehensive update. It's worth noting that it's not necessary to run `script/update` every time a vagrant box is restarted - it basically just loads data, rather than updating the code or dependencies. 

Fixes #42
Fixes #36

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ script/server
 
 The site will be visible at <http://localhost:8000>.
 
-# Get and preprocess data (includes setting up a Solr text index for the plan document text)
+# Get, preprocess and load council, plan, and emissions data (includes setting up a Solr text index for the plan document text)
 
 ```
 script/update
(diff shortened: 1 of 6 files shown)


zarino (approved):
Can confirm `script/update` now skips the downloading steps unless `--all` is requested 👍 

Although I got an exception at the end – not sure whether it was caused by this change, or by my local `data/` cache being quite old?

Screenshot 2020-11-09 at 15 34 29

PR #50, opened 2020-11-05
Upgrade to Django 3

sagepe:
This upgrades Django and some associated packages to the most recent version.

### DIFF
--- a/.github/workflows/django.yml
+++ b/.github/workflows/django.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.5]
+        python-version: [3.7]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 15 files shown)


struan:
I think this is ready for a review, unless anyone has any other changes they think should happen while we're here?

struan:
#50 upgrades django and python to the latest versions, along with a few other packages as required along the way.

I've also included the two version bumps that github auto generated pulls for.

Is there anything else we should be updating while we do this?

sagepe:
Is it worth bumping Haystack? I see that support for Django 3.2 was "officially" added in [3.1.0](https://github.com/django-haystack/django-haystack/releases/tag/v3.1.0) (we've got 3.0.0), but a quick look suggests that this was really only adding testing to CI for this version (although I've not done a full review of that release).

struan:
FWIW, it didn't seem worth updating haystack as there weren't any changes related to solr.

PR #51, opened 2020-11-06
Add simple related councils using MapIt's touch API

### LINKED ISSUE #40: Add some 'related councils' function

crowbot:
Fixes #40.

When running the script with 1 second delays, I hit a MapIt rate limiter error message, despite not being at the limit of the API key - @dracos is there some other limit I should be aware of? I've raised the delays in the meantime. 
Screenshot 2020-11-10 at 19 08 14


dracos:
I don't know of any, I can't see any limiting in the logs. What was the error message, do you know the IP address requests came from/an example request that gave the error?

dracos:
Yeah, so there's nothing limit based in the varnish logs on starling that I can see, which should AFAIK log everything, API key limits, rate limits, what have you. So not sure what it could have been, sorry.

zarino (approved):
👍 Looks good. I’ve added a fixup commit that adds a bit of spacing above the table, and simplifies the table heading.

PR #52, opened 2020-11-09
"Locate me automatically" link, as alternative to postcode search

### LINKED ISSUE #41: Add lat/lon to council function
To support 'Use my location' on postcode search

zarino:
Couldn’t resist.

Fixes #41.

While I was there, I also DRYed up the location search form, so that the same partial is now shared between the homepage and council disambiguation page.

Screenshot 2020-11-09 at 18 10 14

### DIFF
--- a/caps/mapit.py
+++ b/caps/mapit.py
@@ -30,6 +30,7 @@ class MapIt(object):
     # MTD (Metropolitan district)
     # UTA (Unitary authority)
     touches_url = '%s/area/%s/intersects?type=CTY,COI,DIS,LBO,LGD,MTD,UTA&api_key=%s'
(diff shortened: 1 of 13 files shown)


crowbot (approved):
Looks good! It's slow enough that it'd be nice to have some indication of something happening before you get redirected, if that's possible. 

When this gets merge, we should add the API key to the MapIt call, as in https://github.com/mysociety/caps/pull/51

In testing it, I've just noticed that submitting the postcode form empty causes an error, so I've added a commit to fix that.

zarino:
@crowbot I’ve added a loading spinner while the geolocation API is being called, and I’ve rebased onto the lastest `master`, pulling in the MapIt API key change.

crowbot (approved):
Perfect!

PR #58, opened 2020-11-10
Styling for text search results page

### LINKED ISSUE #57: Styling for text search results page
eg: http://localhost:8000/search/?q=electric+taxi

zarino:
Includes "popular searches" list on homepage search widget.

A few things that would be nice to add in future:

1. Display what _type_ of document we’re linking to – or even better, the _title_ of the document (rather than just "plan document").
1. "x–y of z results" message (rather than just "z results") so you can see how many pages of results there are.
1. numbered pagination buttons (rather than just "previous" / "next" buttons).

Fixes #57.

Screenshot_2020-11-11 Home(1)


caps/views.py
```diff
@@ -20,6 +20,12 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['total_councils'] = Council.objects.all().count()
         context['percent_councils_with_plan'] = Council.percent_with_plan()
+        context['search_suggestions'] = [
+            '"electric taxis"',
+            '"green homes grant"',
+            'ground source heat pumps',
```
crowbot, line 26: These last two need double quotes inside the single quotes like the first two.

crowbot (changes requested):
Looks great! One issue to resolve inline.

caps/views.py
zarino: I left the other two without double quotes because they seemed to bring up fairly sensible results without them. Also hints to people that they can use or not use double quotes, just like a Google search, but maybe that’s a bit subtle.

crowbot: They start out as good results but then do degenerate into the individual words in different places in the documents. I think probably better to keep them in quotes, or add in an example where we don't expect the words to appear together as part of a phrase.

zarino:
@LouiseMySociety I‘ve put quotes around all four search suggestions, and removed the "popular" wording.

crowbot (approved):

zarino:
Don’t know what‘s wrong with me today, I keep tab-completing to the wrong Louise!!

PR #61, opened 2020-11-11
Import local configuration overrides after setting default for ALLOWED_HOSTS

### LINKED ISSUE #60: ALLOWED_HOSTS should be configurable
Currently local configuration overrides are imported before setting `ALLOWED_HOSTS` in `proj/settings.py`, this should be configurable.

sagepe:
This allows us to control this in our configuration templates. Fixes #60.


sagepe:
Closing as this will be included in #62.

PR #62, opened 2020-11-11
Docker integration

sagepe:
This adds support for using Docker in production. Deploys once this is merged, we can made a few subsequent changes to use docker in development as well as ensuring that the master branch gets automatically deployed.

### DIFF
--- a/conf/config.py.docker
+++ b/conf/config.py.docker
@@ -1,0 +1,16 @@
+import os
+SECRET_KEY = os.environ.get("SECRET_KEY", "xx")
+SASSC_LOCATION  = os.environ.get("SASSC_LOCATION", "sassc")
+
+REPOSITORY_DB_HOST = os.environ.get("REPOSITORY_DB_HOST", "postgres")
+REPOSITORY_DB_PORT = os.environ.get("REPOSITORY_DB_PORT", "5432")
+REPOSITORY_DB_USER = os.environ.get("REPOSITORY_DB_USER", "caps")
+REPOSITORY_DB_NAME = os.environ.get("REPOSITORY_DB_NAME", "caps")
+REPOSITORY_DB_PASS = os.environ.get("REPOSITORY_DB_PASS", "caps")
+
+MAPIT_URL = 'https://mapit.mysociety.org'
+
+CACHE_FILE = 'data/cache'
+
+SOLR_URL = os.environ.get("SOLR_URL", "http://solr:8983/solr/caps")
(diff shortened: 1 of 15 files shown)


conf/config.py.docker
```diff
@@ -0,0 +1,16 @@
...
+REPOSITORY_DB_USER = os.environ.get("REPOSITORY_DB_USER", "caps")
+REPOSITORY_DB_NAME = os.environ.get("REPOSITORY_DB_NAME", "caps")
+REPOSITORY_DB_PASS = os.environ.get("REPOSITORY_DB_PASS", "caps")
+
+MAPIT_URL = 'https://mapit.mysociety.org'
+
+CACHE_FILE = 'data/cache'
+
+SOLR_URL = os.environ.get("SOLR_URL", "http://solr:8983/solr/caps")
+SOLR_ADMIN_URL = os.environ.get("SOLR_ADMIN_URL", "http://solr:8983/solr/admin/cores")
```
crowbot, line 18: The `master` branch now also has `MAPIT_API_KEY` defined

crowbot (approved):
I've read through all this and it looks sensible, but I can't say that I know enough about Docker to check that it's right from reading the code - although the presence of data.climateemergency.uk suggests it is 🎉 . I can build it with `docker build .`, should I be able to run it in docker locally with something like `docker run --publish 8001:8000 --detach  --name caps2 [id of the image]? 

I've noted inline that we've added another config variable to master that will need to be added here too.

conf/config.py.docker
sagepe: I'll get that rebased in. Do we have an API key to use?

sagepe:
> I can build it with docker build ., should I be able to run it in docker locally with something like docker run --publish 8001:8000 --detach --name caps [id of the image] and access it in a browser?

This reminds me that I haven't added any documentation - I'll make some initial additions. 

The script can be built with `script/build`; the image won't run on its own as it needs other containers to make sense - I'll add some instructions for testing.

conf/config.py.docker
crowbot: I've been using a key associated to my mySociety email address that has a subsciption at the 'up to 10,000 calls' level. Do we have a more generic mySociety account for our own usage?

sagepe: As noted elsewhere - we have whitelisted the internal IP so won't need a key in production. I also have a key of my own for testing.

PR #63, opened 2020-11-11
Individual council page improvements

zarino:
Tidying up the council page.

Not massively happy with how this section below the page heading is looking, but it’s better than nothing for the demo.

Screenshot_2020-11-11 Reading Borough Council

@crowbot Could you work out how to feed real data into the two bits marked up as `<!-- TODO -->` in the HTML template?


zarino:
I love the idea of having multiple sparklines – although in this case we’re presenting the same data in each, just divided by either the population or the area (which stays roughly the same each year) so the lines look mostly identical. Which is a bit, meh.

That said, one thing I like about presenting _these three_ figures in particular, for the demo, is it’s a good opportunity to ask people _which_ figure is actually most useful to them. Total, per capita, or per km. Because it would be nice to narrow down to fewer stats, especially if we want to display these figures in the tables on the council list page and the "nearby councils" section of the individual council page.

Also, if we say "in scope of influence" on one chart, I guess we should say it on all of them.

crowbot (approved):
👍

PR #66, opened 2020-11-17
Only update related councils on a full update

crowbot:


sagepe (commented):
Looks good to me - I tested this locally, rebased against current master (it rebased cleanly) and it ran fine, and quickly!

sagepe (approved):

PR #71, opened 2020-12-03
Add google analytics

crowbot:
`client_storage` is turned off to prevent a tracking cookie being used

### DIFF
--- a/caps/context_processors.py
+++ b/caps/context_processors.py
@@ -1,0 +1,4 @@
+from django.conf import settings
+
+def analytics(request):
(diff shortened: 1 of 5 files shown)


zarino (approved):
Looks good to me! :+1:

sagepe:
This will need to have an addition to `conf/config.py-docker`, similar to [line 14](https://github.com/mysociety/caps/blob/f6a95da9899b6b95477d823cf2c51bb4313946f1/conf/config.py.docker#L14).

`GOOGLE_ANALYTICS` will be in the environment with the correct contents.

Issue #70, opened 2020-11-19
Update attributes on an existing plan_document on script/update 


crowbot:
At the moment, if you change, for example, the plan type in the spreadsheet, but the plan has already been loaded into the database, the plan type doesn't change, but it should do. I think we probably can fix this using https://docs.djangoproject.com/en/3.1/ref/models/querysets/#update-or-create rather than get_or-create.

struan:
Fixed in 3894654

PR #81, opened 2021-05-26
Confirm import changes

### LINKED ISSUE #68: Do some kind of change summary before updating the database
As a protection against vandalism, removal or corruption of data by error.

struan:
Update the plan import management command to output information about what is about to change, and only commit if a `--commit_changes` option is used.

This also updates plans if the data in the spreadsheet has changed.

fixes #68 and #70


caps/management/commands/import_plans.py
```diff
@@ -15,8 +15,67 @@
 class Command(BaseCommand):
     help = 'Imports data from data/plans into the database'
 
+    verbosity = 1
```
crowbot, line 18: What does this do?

caps/management/commands/import_plans.py
```diff
@@ -15,8 +15,67 @@
...
+            if not plan_exists:
+                plan_add_count += 1
+                self.print_change("adding new plan for %s", row['council'], verbosity=2)
+            else:
+                plan = PlanDocument.objects.get(
+                    url=row['url']
+                )
+                diffs = 0
+                for key, value in self.get_plan_defaults_from_row(row).items():
+                    if getattr(plan, key) != value:
```
crowbot, line 63: This was producing a high number of expected updates for me because the `start_year` and `end_year` attributes were being returned from the database as `int` types but from the `get_plan_defaults_from_row` function as `string` type. One possible fix would be the following change in `PlanDocument.start_and_end_year_from_time_period`:

```diff
-    return (match.group('start_year'), match.group('end_year'))
+    return (int(match.group('start_year')), int(match.group('end_year')))
```

crowbot (changes requested):
Looks really great, thanks! 🎉 

Sorry that the code you're working with is a bit rudimentary. I've added some comments inline on a couple of things that I _think_ need fixing before this gets merged.  

In reviewing the PR, I got hung up by the fact that the data in the current spreadsheet had a few instances where the same plan url appeared twice for a council, which made it look like some updates weren't being applied (the record was being updated based on the first spreadsheet row, then updated again based on the second spreadsheet row). I'll ticket that separately.

caps/management/commands/import_plans.py
struan: Demonstrates that I didn't understand what was going on with the manage command default args :|

crowbot (approved):
Looks great, thanks :shipit:

PR #84, opened 2021-06-02
Bump various dependencies

sagepe:
This updates requests, urllib3, pillow, django and pyyaml. The latter four were flagged individually by Dependabot; the urllib3 tests failed due to the need to update requests as a dependency.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,6 @@ django-pipeline==1.6.14
 et-xmlfile==1.0.1
 feedparser==5.2.1
 future-fstrings==1.2.0
-futures==3.0.5
 gunicorn==20.0.4
 html5lib==1.1
 idna==2.10
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ chardet==3.0.4
 csscompressor==0.9.4
 defusedxml==0.6.0
 diff-match-patch==20200713
-Django==2.2.8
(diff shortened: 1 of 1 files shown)


struan (commented):
When I try `docker compose -f development.yml build` I get an error: `ERROR: No matching distribution found for futures==3.0.5`

sagepe:
> When I try `docker compose -f development.yml build` I get an error: `ERROR: No matching distribution found for futures==3.0.5`

Odd. My first rebuild worked, so I was initially a bit confused. Then I checked and realised I had local old copies of the base images `python:3.7` and `python:3.7-slim` cached so I pulled the latest versions, rebuilt and was able to reproduce the error.

Not sure why though, this built using 3.7.10 in CI, the same version used in the latest Docker images. Plus, when I checked the `futures` package, it seems it shouldn't be needed anyway and according to [the project description](https://pypi.org/project/futures/3.3.0/#description) shouldn't work at all with Python3.

> It does not work on Python 3 due to Python 2 syntax being used in the codebase. Python 3 users should not attempt to install it, since the package is already included in the standard library.

🤷🏻

So I've removed `futures` from `requirement.txt`. Running a `docker-compose -f development.yml build --no-cache` now completes successfully. I've just pushed that commit into this branch, so we'll see if it passes CI.

struan (approved):
yay, that all worked

PR #90, opened 2021-07-06
plan_link_click analytics event created

ajparsons:
This PR solves a problem where the GA `file_download` event does not always make it easy to link back to the council the file is associated with. 

This explicitly tags up links in the council and search page with the council slug, and has a `gtag` function that catches these links and triggers a `plan_link_click` event.

### DIFF
--- a/caps/static/js/main.js
+++ b/caps/static/js/main.js
@@ -121,4 +121,63 @@ $(function() {
         );
     });
 
+    $('a[data-plan-id]').on("click", function(e) {
+        var link = $(this);
+        var link_url = link.attr('href');
+        var link_text = link.text();
+        var link_ext = "." + link_url.split('.').pop();
+        var file_name = link_url.split('/').pop();
+        var council_slug = link.data("council-slug");
+        var plan_id = link.data("plan-id");
+
+        // Set up callback to open link in current window,
+        // if user hasn't initiated click with modifier keys.
+        var callback;
+        if (
+            e.ctrlKey ||
+            e.shiftKey ||
+            e.metaKey ||
+            (e.button && e.button == 1) // middle mouse button
+        ){
+            callback = function(){}
+        } else {
+            e.preventDefault();
+            callback = function(){
+                window.location.href = link_url;
+            }
+        }
+
+        trackEvent('plan_link_click', {
+            file_extension: link_ext,
+            file_name: file_name,
+            link_url: link_url,
+            link_text: link_text,
+            council_slug: council_slug,
+            plan_id: plan_id
+        }).done(callback);
+    });
 });
+
+var trackEvent = function(eventName, params) {
+    // We'll return a promise, and resolve it when either Gtag handles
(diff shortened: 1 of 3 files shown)


caps/static/js/main.js
```diff
@@ -121,4 +121,25 @@ $(function() {
...
+    $("a").click(function(e) {
+        link = $(this);
+        if (link.data("plan-id")) {
+            link_url = link.attr('href');
+            link_text = link.text();
+            link_ext = "." + link_url.split('.').pop();
+            file_name = link_url.split('/').pop();
+            council_slug = link.data("council-slug");
+            plan_id = link.data("plan-id");
+            gtag('event', 'plan_link_click', {
```
zarino, line 133: I _think_ there’ll be a race condition here, where the communication with Google might not have a chance to happen before the default action of the link kicks in and the page reloads.

The gtag docs include an example of passing an `event_callback` parameter along with the other event parameters, to specify code that should run _after_ the event has been successfully logged:

https://developers.google.com/analytics/devguides/collection/gtagjs/sending-data#implement_event_callback_functions

Back in the universal analytics days, we had a helper library for doing all of this: https://github.com/mysociety/analytics#analyticstracklinkclicke-params – maybe you could update the helper library to use `gtag()` if it’s available, and then we could use that?

caps/static/js/main.js
```diff
@@ -121,4 +121,25 @@ $(function() {
         );
     });
 
+    $("a").click(function(e) {
```
zarino, line 124: A stylistic thing, but you probably want `.on('click', …)` here, rather than `.click()`.

I was originally going to say, binding to _all_ anchor elements on the site feels wasteful, but I bet it actually makes no difference in reality. If you wanted to only bind to links with a `plan-id` attribute, you could do it by selecting: `$('a[data-plan-id]')`. Then you wouldn’t need the `if` inside the callback.

zarino (changes requested):
Have left a few comments. I haven’t tried it out to verify the race condition, but I know it definitely _used_ to be the case, and so probably worth defending against it anyway, especially as we’re likely to use link tracking in a number of places, we should get it right.

ajparsons:
So I've roughly got a version that does everything other than catch `right-click > new tab` or middle click in chrome (this seems either impossible or increasingly fiddly). 

This seems fine to start with, but the other option is to change the link to an internal url, which then sends the event via [measurement protocol](https://developers.google.com/analytics/devguides/collection/protocol/ga4) before redirecting. This moves analytics into django, and means all clicks get download events. But might also raise issues with caching?

zarino:
@ajparsons I’ve added a fixup commit that:

* Avoids the special "have we already run the callback?" handling, by using a jQuery Deferred object (promise) which can only be resolved once, so problem solved.
* Declares variables with `var` to avoid accidentally creating global variables.
* Renames `safeLinkTag` to `trackEvent`, and moves it outside the jQuery.ready callback, just in case we want to call track events from outside that callback in the future.

PR #91, opened 2021-07-06
Adds link and email capture for survey

ajparsons:
This adds a link and an email capture form for a survey to a council page. 

The link goes to an Alchemer survey, the email box goes to a Mailchimp audience which will automatically email a copy of the survey four days after filling out the box before unsubscribing the user. 

It also adds a quick link to the survey in the 'about' page as the Alpha feedback banner takes you there.


zarino (approved):
This was a tricky one to style up, given the survey content was so large and the rest of the page so sparse. I’ve added a commit that shortens the survey text a little (hope you don’t mind?) and displays the survey box side-by-side with the plan(s):

It also adds a button for the survey, rather than hiding the link inside paragraph text, and makes the survey open in a new window.

Screenshot 2021-07-06 at 13-05-01 Liverpool City Council

The layout of this page will hopefully be changing soon, so while this isn’t perfect, I think it’s probably good enough for the short time it’ll be like this.

ajparsons:
This is great! Let me wait for @Rebeccarumbul 's feedback on the survey and then we can deploy it.

PR #93, opened 2021-07-12
Django intermediary between survey and mailchimp

ajparsons:
Obviously rewrite the JS bit as you like, but in principle creates a local hook for adding an email to the survey queue.

You need to add MAILCHIMP_KEY to the .env file - you can source the CAPS key from https://us9.admin.mailchimp.com/account/api/

(Not sure where we'd need to add that to deploy properly)


zarino:
Thanks. I’ve cherry-picked your changes into https://github.com/mysociety/caps/pull/92

PR #98, opened 2021-08-02
Allow location search by council name, not just postcode

### LINKED ISSUE #88: Homepage could have lookup by council name
Currently the homepage does lookup by postcode, but users may also want to look up a council they don't know the postcode for.

For instance, if you're a council worker who lives out of area, you might not be able to simply find your own council. Or if you have been told to look at a specific plan, have to click off page to find the link. 

IMD maps website has a version with two textboxes and an auto-complete on the 'short' version of the council name https://research.mysociety.org/sites/imd2019/

zarino:
Postcode search inputs on homepage and location search results page now display an autocomplete list of council names when you type three or more characters, if JavaScript is available.

Screenshot 2021-08-02 at 14 23 00

Selecting a council from the list immediately navigates to that council’s page, rather than submitting the form (since the form can only handle postcodes right now).

If the postcode search input is inside an interstitial-triggering form (which it is, right now), the interstitial-triggering logic is run before the page location is changed.

Fixes #88.


caps/templates/includes/location-search-autocomplete.html
```diff
@@ -0,0 +1,32 @@
...
+);
+document.addEventListener('awesomplete-select', function(selection){
+    selection.preventDefault();
+    if ( $pc.parents('form[data-show-interstitial]').length ) {
+        handleInterstitialTrigger();
+    }
+    window.location.href = councils[selection.text];
+});
+
+var $label = $('label[for="pc"]');
```
struan, line 28: This highlights the bug that the label for the phrase search has an incorrect for attribute so also gets its label changed.

struan (changes requested):
One tiny fix required.

We should probably also have a ticket to make this work without JS at some point.

caps/templates/includes/location-search-autocomplete.html
zarino: Whoa, nice spot! How did I miss that!!

I’ve amended the commit, to change the text search label to `for="q"`, which it should have been all along.

PR #99, opened 2021-08-05
delete plans no longer in spreadsheet from database

### LINKED ISSUE #67: Remove obsolete plan documents when doing a quick update of the database.
At the moment, documents that were previously associated with a council, but which have been removed from the spreadsheet are still retained in the database.

struan:
adds output to say how many plans are being removed and also if a
council is having all their plans removed.

Fixes #67


caps/management/commands/import_plans.py
```diff
@@ -75,12 +85,48 @@ def get_changes(self):
                         plan_update_count += 1
                         self.print_change("updating plan for %s", row['council'], verbosity=2)
 
+        plans_to_delete = {}
+        plans_to_delete_count = 0
+        for council_code in plans_to_import.keys():
+            councils_in_sheet.update([council_code])
```
crowbot, line 91: Does this line do anything given the similar lines earlier in this function?

caps/management/commands/import_plans.py
```diff
@@ -75,12 +85,48 @@ def get_changes(self):
...
+            council = Council.objects.get(
+                gss_code = council_code
+            )
+            plans = PlanDocument.objects.filter(
+                council=council,
+            ).exclude(
+                url__in=plans_to_import[council_code]
+            )
+            cp = PlanDocument.objects.filter(council=council, url__in=plans_to_import[council_code]).count();
```
crowbot, line 100: Not used again?

caps/management/commands/import_plans.py
```diff
@@ -75,12 +85,48 @@ def get_changes(self):
...
+            for plan in plans:
+                plans_to_delete_count += 1
+                council_plans = plans_to_delete.get(council_code, set())
+                council_plans.update([plan.url])
+                self.print_change("deleting plan for %s", council.name, verbosity=2)
+                plans_to_delete[council_code] = council_plans
+
+        councils_removed_plans = PlanDocument.objects.exclude(
```
crowbot, line 108: I know it's super verbose, but I think `plans_from_removed_councils` might be clearer here, as we're juggling a few similar variable names?

crowbot: Bit out of scope given the ticket, but perhaps we should remove those councils here too.

crowbot (changes requested):
Looks good to me! A few comments inline

caps/management/commands/import_plans.py
struan: yeah, I wasn't sure about removing them entirely. I guess the only time they are likely to be removed is for things like e.g. Somerset going unitary as we include a line with no plan if they have no plan, in which case yes, removing them is probably the best thing as they don't exist any more.

crowbot: Yeah, that's the situation where removing councils has come up so far.

struan: fwiw, the code does now remove them.

crowbot (approved):
Looks good to me - one comment inline. It occurs to me that it might be worthwhile when you deploy this to do a dry run on the production server so that we can see and sense check the changes this will make to the live data before they're made?

caps/management/commands/import_plans.py
```diff
@@ -104,18 +109,37 @@ def get_changes(self):
...
+        # if a council isn't in the sheet we should remove it entirely from the database
+        councils_to_remove = Council.objects.exclude(
+            gss_code__in=councils_in_sheet
+        )
+        councils_to_remove_count = councils_to_remove.count()
+
+        # if a council is going to be removed completely then exclude it from the
+        # list of councils where we're going to remove the plans
+        councils_with_plans_to_remove = set([ council.gss_code for council in councils_to_remove ])
+        councils_with_plan_in_sheet = councils_with_plan_in_sheet.symmetric_difference(councils_with_plans_to_remove)
```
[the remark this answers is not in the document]
struan: er, yeah, it turns out I can just do an update which has the same effect but is more obvious.

PR #100, opened 2021-08-05
management command to generate a zip file of all plans

struan:


zarino:
Screenshot 2021-08-11 at 11 29 01

I wonder whether we should give people a heads up that it (I assume?) is a very large file? Is that possible?

zarino (approved):
I’ve not tested out the management command, but the front-end changes look good 👍

caps/templates/about.html
```diff
@@ -18,6 +18,8 @@ <h2 class="mt-5">Data sources</h2>
 
     <p>Links to council climate action plans were crowdsourced by an army of volunteers <a href="https://docs.google.com/spreadsheets/d/1tEnjJRaWsdXtCkMwA25-ZZ8D75zAY6c2GOOeUchZsnU/edit">in this spreadsheet</a>. If you find a plan document that we’ve missed, <a href="https://docs.google.com/document/d/1oGFkzhZkTCfEBUeJIdMb9aOfuqPoCHkUMTJxROsRRow/edit">read our guide</a> on what we consider a climate action plan and how to add one to our spreadsheet.</p>
 
+    <p>You can <a href="{{ MEDIA_URL }}data/plans/plans.zip">download a zip file of all the plans</a>.</p>
```
crowbot, line 21: Might be worth adding here something to convey that you also get the plans.csv file, which tells you more about each plan, and has the council gss_code, which will allow this data to be linked to other data.

crowbot (approved):
Looks good to me! 🥳  One comment inline about being more detailed on the about page in describing what you get in the download. 

The HTML format files I looked at are pretty useless in the download as there's no style files and the links in them are relative, but I think given that we include `plans.csv` to decode what they are (mainly overview pages from the councils' websites), it's worth keeping them in for completeness. I do recall that perhaps at least one council has made their strategy available as HTML not PDF.

struan:
I was mostly assuming that people would be using this to push it into their own indexing/analysis tools so from that perspective the HTML was still potentially useful. This is based on no evidence whatsoever though :)

caps/templates/about.html
struan: How about "This includes a CSV file (plans.csv) with details and sources for all the included files, along with information like GSS codes to enable linking to other data."

crowbot: 👍 perfect

PR #104, opened 2021-08-11
record and display recent search terms

struan:


zarino:
@struan I‘ve added a fixup commit that removes the 4 recent searches (not sure they really add much, when we’ve got popular searches instead), and another fixup commit that tidies up the styling of the popular searches too. Now looks like:

Screenshot 2021-08-18 at 16 23 56

zarino (approved):
UI looks fine to me. I seem to have got my DB into some wierd state where I get errors about `column "inorganic" of relation "caps_savedsearch" does not exist` but `script/migrate` doesn’t change anything, so unable to test out the functionality. Maybe @crowbot will have better luck.

crowbot (approved):
Looks good to me!

struan:
@zarino I think that might be because I did some refactoring of the migrations. If you run

`migrate caps 0026_auto_20210406_1835`

and then the normal migrate again it should fix it, I think.

PR #109, opened 2021-08-17
WIP: council promises

struan:
storing, importing and displaying council promises on becoming carbon neutral


struan:
@zarino do you want to have a look at how I've done the presentation on this?

struan:
Also, I wonder if we want to be able to distinguish between "not checked" and "could not find a statement"

zarino:
@struan looks good. I’ve added a commit that tidies up the vertical spacing a bit, and swaps the yellow for a more legible orange.

Do you think we could get a link to the source document, inside each `<cite>`? I tried a naïve `<a href="{{ promise.source_url }}">…</a>` but it came out empty. I must have missed something.

Also, in the situation where there’s no promises, I wonder whether the wording should be "promises from this council" rather than "promises _for_ this council"? After all, it’s the council making the promises!

> Also, I wonder if we want to be able to distinguish between "not checked" and "could not find a statement"

Originally I assumed we wouldn’t go live with any of this data until we’d at least done a first crowdsourcing pass checking each website. But maybe there’s value in publishing promises as we find them. @LouiseMySociety @emilyk383 any thoughts?

Either way, it may still be useful to distinguish "We couldn’t find any climate promises for this council" vs "We haven’t yet checked for climate promises for this council" – it’s not unheard of for councils to merge / split, so even after our initial sweep, there could be new councils that pop up and need checking, which we might not do immediately. Guess it depends how hard it’d be to implement a flag like this, @struan?

struan:
I can do the cite thing and yes, "from" makes more sense than "for".

The flag is easy, not sure how we'd do it in the spreadsheet. I suspect adding a No Promise to the scope would be the easiest even if it's not entirely the obvious place. And then we can add display a No in the net zero by column

struan:
I've added an option to the import to handle a Scope of "No promise" so that could do with adding to the spreadsheet. If the scope is that it created a promise but sets has_promise to False.

zarino:
> I've added an option to the import to handle a Scope of "No promise" so that could do with adding to the spreadsheet.

Cool! I’ve added "No promise" as an option for the data validation on column B of the spreadsheet, and updated that cell for each of the councils already in the spreadsheet (councils beginning with the letter A).

emilyk383:
Wrt publishing promises as we find them vs checking all sites first... I can see why the latter would be ideal but wonder if we'll be able to find volunteer capacity while the scoring is going on. Could we put something on the front page about promises being a 'new feature' - encouraging visitors to check out and tell us whether their Council has made a promise or not? (So yes, I'd favour publish the ones we know about now, and let's see how fast it moves forward organically)

zarino:
> wonder if we'll be able to find volunteer capacity while the scoring is going on

Could talk to CEUK about it! I have a feeling they’ll be worried that the volunteers already have enough on their plate, and spotting and sourcing the "promise" (which might not be in the Action Plan) might be too much. But we could see!

struan:
I've rebased this a chunk to make it fewer, more sensible commits. you might need to undo and redo migrations if you've already done anything with it.

emilyk383:
> Could talk to CEUK about it! I have a feeling they’ll be worried that the volunteers already have enough on their plate, and spotting and sourcing the "promise" (which might not be in the Action Plan) might be too much. But we could see!

Do you want to bring it up or shall I? :D

zarino:
>> Do you want to bring it up or shall I? :D
> 
> We already asked them this I think

Yeah, just discussing with @emilyk383 now. We’ll have a think 👍

caps/templates/council_detail.html
```diff
@@ -48,23 +48,29 @@ <h2>Other nearby councils</h2>
...
             </div>
             <div class="card-body">
-                <blockquote class="blockquote mt-n2">
-                    <p class="mb-0">“carbon neutral for our direct emissions by 2025”</p>
-                    <footer class="blockquote-footer"><cite title="Source Title">Mayor’s Climate Emergency Action Plan, 2019</cite></footer>
-                </blockquote>
-                <blockquote class="blockquote mb-0">
-                    <p class="mb-0">“carbon neutral and climate reslilent by 2030”</p>
-                    <footer class="blockquote-footer"><cite title="Source Title">Bristol One City Climate Policy, 2020</cite></footer>
+                {% for promise in council.promise_set.all %}
```
zarino, line 56: I think the logic for this section is wrong. Eg: when a council has both a promise and a "no promise" line in the source spreadsheet (like Allerdale right now):

image

I tried fixing it but couldn’t get the right combination of booleans in the `{% if %}` tag!

As mentioned in Slack, I think it should go something like:

```
{% if promise_set_is_not_empty and first_promise_isnt_explicit_no_promise %}
  {% for each promise in promise_set %}
    {# … print the blockquote etc #}
  {% endfor %}
{% else %}
    <p>
        We haven’t yet found any carbon zero target dates for this council.
    </p>
    <p class="text-muted mb-0">
        Have you found one? <a href="{% url 'about' %}">Let us know!</a>
    </p>
{% endif %}
```

Note the HTML in the `{% else %}` tag is different to currently, and is also the same regardless of whether the we’ve checked the council for promises or not.

Eventually we’ll want to link to a Google Form or something for people to suggest improvements, but for now, a link to the About page (same as the "Alpha" banner at the bottom of the site) is better than nothing.

zarino (approved):
Just one issue with the display of (non-)promises on the individual council page.

I’ve added two commits, one to slightly tweak the wording on the council list page, and one to move the filters on that page into a sidebar, rather than being along the top (since we’ll likely want to add more filters soon anyway).

struan:
FWIW, the current state with this is we're trying to decide what data about council promises to display.

Obviously the target date, statement etc is a given. What's not clear is if we expose when we can't find a council statement versus not having checked for one. The spreadsheet allows for this distinction to be collected but it's not clear if we want to highlight it.

There is also the question of what to display in the data table when a council has a statement but hasn't set a date. At the moment this is treated the same as not having a statement but it's not quite the same.

zarino:
> What's not clear is if we expose when we can't find a council statement versus not having checked for one.

ie: Do we want to differentiate between "we looked but we couldn’t find anything" and "we haven’t looked yet".

Sorry I’ve been back and forth on this a bit. My thinking earlier this week was that we can’t be sure the lack of a promise in our dataset necessarily means the council has never made a promise, so we should hedge our bets with a "we couldn’t find"-style message even if we _have_ checked, which led me to think, for simplicity, lets just combine to the two failure cases (which, from the look of CEUK’s spreadsheet is only ~12% of councils) into the same "we couldn’t find" message.

But, thinking more about it, the original impetus for displaying promises was:

1. People want to use the presence of a carbon zero target date to answer the question: "does this council care about the climate emergency at all?"
2. People want to use the closeness/distance of the target date to answer the question: "is this council handling its climate response with urgency?"

So it _does_ feel important that we show the difference between: A) a promise being absent because we looked and couldn’t find one; and B) a promise being absent because we just haven’t looked yet. Visitors could rightly read the former as a signal that the council isn’t taking the climate emergency seriously, but they shouldn’t draw any conclusions from the latter – chances are the council _has_ a promise, we just haven’t added it to our site yet.

Thing is, if we go back to something like this:

```html
{% for promise in promises %}
    <blockquote class="blockquote">
        <p class="mb-0">“{{ promise.text }}”</p>
        <footer class="blockquote-footer"><cite title="{{ promise.source_name }}"><a href="{{ promise.source }}">{{ promise.source_name }}</a></cite></footer>
    </blockquote>
{% empty %}
  {% if checked_but_not_found %}
    <p>This council hasn’t published any net zero target dates.</p>
    <p class="text-muted mb-0">Have you found one? <a href="{% url 'about' %}">Let us know!</a></p>
  {% else %}
    <p><!-- errr… --></p>
    <p class="text-muted mb-0">Have you found one? <a href="{% url 'about' %}">Let us know!</a></p>
  {% endif %}
{% endfor %}
```

…What do we put in that final slot? "We haven’t searched for net zero targets for this council yet"? Looks a bit half-hearted, doesn’t it?

I think I‘d be ok with that, if we knew it was going to be an edge case. (ie: if we hold off on launching this feature until we’ve got a fairly full dataset of promises.)

> There is also the question of what to display in the data table when a council has a statement but hasn't set a date. At the moment this is treated the same as not having a statement but it's not quite the same.

I think as long as the column in the table is titled "Net Zero target" (or maybe "Net Zero target date" to make it super clear?) it’ll be fair to leave the cell empty if a council hasn’t given a target date.

Would be interested to know @crowbot’s thoughts on these two points as well.

emilyk383:
Could we say 'We haven’t found net zero targets for this council yet' in that last slot? Also - planning to put a proposal together for the admin we want to outsource together today, so we *might* be able to get these relatively quickly. 

The other thing that occurred to me was to ask the Councils to tell us this when we write to them with the right to reply. I didn't realise, but apparently not a lot of folk know that the action plan spreadsheet is open data that anyone can add to - we could ask them to put it in - then the last box could say 'they haven't told us yet' ??

zarino:
> Could we say 'We haven’t found net zero targets for this council yet' in that last slot?

Well this is the million dollar question – that wording was my default position, but it implies that we’ve at least _tried_ to find a target, which isn’t the case.

> we could ask them to put it in - then the last box could say 'they haven't told us yet' ??

Mmmmm, evil genius. Yeah, I could imagine something along the lines of "This council hasn’t provided any net zero target dates" once we’ve got an established process for getting councils to submit their own data.

emilyk383:
I've been having super-fun (not even joking) putting [this admin outsourcing](https://docs.google.com/document/d/1jpJdTPj8uem2e8E2fUrQhTLcBqNR6sPpqtE0LxnnVxA/edit) doc together. I just emailed it to you and Louise but actually prob should have posted here... The wording issue might be resolved if displaying Council promises CAN wait until we've the majority under our belts / we've at least asked? How quickly that can happen depends on whether we go with carbon neutral or net zero...

struan:
I think for the third option could we say, at least initially, something like "we're in the process of checking all councils for net zero target dates and we've not checked this one yet." only with better words.

zarino:
PS thanks to @emilyk383 for pointing out (somewhere? lost the reference now!) that carbon neutral and net zero aren‘t the same thing, so we should be careful about using the terms.

From a quick analysis of the NAO dataset (which contains exact quotes of council promises) here’s how often the different terms appear:

| % of promises | term |
| --- | --- |
| 50% | "carbon neutral" |
| 24% | "net zero carbon" or "net-zero carbon" or "net carbon zero" |
| 14% | "net zero" or "net-zero" (without "carbon") |
| 3% | "zero carbon" (without "net") |

struan:
FWIW it's worth I've gone with

> We’re in the process of checking all councils for climate promises and we’ve not got to this one yet so can't say if they've made a promise or not.

I am not clear what we've decided on the "Net Zero" vs "Carbon Neutral" debate. Or have we decided to just go for "Net Zero" and let people read the statement to work it out? One other option is "Carbon target date" and then a little ? rollover style thing that explain it mean either Net Zero or Carbon Neutral and you have to read the statement? I realise that adding our own special terminology is possibly not helpful.

zarino:
> I am not clear what we've decided on the "Net Zero" vs "Carbon Neutral" debate. 

I think "climate promises" as you’ve got it, on the council detail page, is good.

For the column header (and filter label, I guess) on the all councils table – it’s tempting to hedge our bets with just "Target date"… but then that’s even more ambiguous. Maybe the best solution for now is just to use the term "carbon neutral" since that appears to be by far the most common type of target date amongst councils, and (from my understanding) net zero is a _superset_ of carbon neutral, so at least we won’t be saying councils have promised to reach the harder target when they haven’t.

zarino (approved):
Looks good. I’ve tweaked the wording again, thanks to @MyfanwyNixon 🙂

Guessing lots of the later commits, from 15th Sept onwards, can be squashed into a single commit?

Think we probably want to hold off on merging this to `master` until we’ve got an idea of how long it’ll be until we’ve collected the required data. If it’s only a few weeks away, we could wait. But if it’ll be longer, maybe we manually populate the dozen or so most popular councils, and deploy this change as-is.

struan:
OK, this is squashed and the two initial commits are now in #132 so will be reviewed separately, and then this rebased once they're merged.

PR #110, opened 2021-08-18
basic read only council api endpoint

### LINKED ISSUE #101: How to deep link to council sites if we only have a council domain?
Hi there,

I'm writing this as an issue for some pointers on how best to deep-link to content on your site, from the own green checking service we make use ourselves in the green web green check tool at the link below:

https://www.thegreenwebfoundation.org/green-web-check

We currently work by doing a DNS look up on a domain, and then finding some information, about the corresponding IP range in use for a given site.

So, for a site like cloudflare in addition to marking it as green, we point to the underlying credentials and any information we have in the context of climate.

See an example below, where we point to their own comms information, and also the numbers and evidence base they are sharing publicly:

https://www.thegreenwebfoundation.org/green-web-check/?url=cloudflare.com

### Doing something like this for public bodies as well as private companies:

We're looking at being able to do something like this, but extend our ability to return information for a "green check" so instead of just showing that hackney is using google as a host, we can check the domain against a list of entities and show richer information.

So for example, any domain that was subdomain of hackney.gov.uk:

https://www.thegreenwebfoundation.org/green-web-check/?url=hackney.gov.uk

We'd be able to also link to this page here:

https://data.climateemergency.uk/councils/london-borough-of-hackney/

We're aiming to define a convention for organisations to disclose information at a domain level called carbon.txt, that can be consumed by anything using our API - we are serving around 7-8 million API requests per day at present, and experimenting with making the data in our API responses richer to people think more holistically about climate action.

But until we have that sorted out, we're looking to rely on some kind of mapping between a domain, and  a useful url to present for users.

### How I think it might work

I see in issue #7  that there that has a list of council website urls, that we could use as a starting point.

It would be really cool to be able to go from: `hackney.gov.uk` 

To an address like this:

https://data.climateemergency.uk/councils/london-borough-of-hackney/

I _can_ see the a link back to hackney there on the https://data.climateemergency.uk page, and this makes me think it might be possible to expose a lookup table.

If that existed, given a council domain as a key, I think we'd be able to link to the relevant page, but I can't see how to generate it without some pointers from someone at Mysociety who knows the data better.

Would something like sharing a JSON or CSV representation of a query along these lines work?

```python
[
  (council.website_url, council.get_absolute_url() )
  for council 
  in Council.objects.all()
]

```

I'm basing that code on reading the Council model at the url below:

https://github.com/mysociety/caps/blob/master/caps/models.py#L19-L69

### Some other background

I'm a big fan of this project - I spoke about why I think it's important, and a good example of using technology to help engage on a societal level at Djangocon EU this year.

The videos of the talk is now on youtube, and the talk is a bit long, so I've just linked to the specific bits on youtube mentioning mysociety and caps:

The first mention of caps and mysociety's work around climate action plans (2-3 mins)
https://youtu.be/quLFLH36Lro?t=1610

The second mention and a call to action around caps (2-3 mins):
https://youtu.be/quLFLH36Lro?t=2821

There's more at the link below, including the entire deck I used if you want to link to any of it:
https://www.thegreenwebfoundation.org/djangocon-eu-2021/

struan:
Uses the Django REST framework to add a basic endpoint for council data

fixes #101

### DIFF
--- a/caps/api/routers.py
+++ b/caps/api/routers.py
@@ -1,0 +1,5 @@
+from caps.api.views import APIView
+from rest_framework import routers
+
+class Router(routers.DefaultRouter):
+    APIRootView = APIView
--- a/caps/api/serializers.py
+++ b/caps/api/serializers.py
@@ -1,0 +1,13 @@
+from caps.models import Council
+from rest_framework import serializers
+
+
+class CouncilSerializer(serializers.HyperlinkedModelSerializer):
+    url = serializers.URLField(source='get_absolute_url')
+    plan_count = serializers.IntegerField()
+    country = serializers.CharField(source='get_country_display')
+    authority_type = serializers.CharField(source='get_authority_type_display')
+
+    class Meta:
+        model = Council
+        fields = ['name', 'url', 'website_url', 'gss_code', 'country', 'authority_type', 'plan_count']
--- a/caps/api/views.py
+++ b/caps/api/views.py
@@ -1,0 +1,27 @@
+# -*- coding: future_fstrings -*-
+from django.db.models import Count
+
+from rest_framework import viewsets, routers
+
+from caps.models import Council
+from caps.api.serializers import CouncilSerializer
+
+class APIView(routers.APIRootView):
+    """
+    API for data about councils and action plans.
+    """
+    pass
+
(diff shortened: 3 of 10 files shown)


crowbot (approved):
Looks good to me! A couple of comments inline - happy for you to merge once those are resolved.

PR #122, opened 2021-09-09
API to search term data

struan:


ajparsons:
Could you add `end_date` as well? Simplifies re-running historical queries.

caps/api/views.py
```diff
@@ -25,3 +25,16 @@ class CouncilViewSet(viewsets.ReadOnlyModelViewSet):
     """
     queryset = Council.objects.annotate(plan_count=Count('plandocument')).all()
     serializer_class = CouncilSerializer
+
+class SearchTermViewSet(viewsets.ReadOnlyModelViewSet):
+    """
+    List of search terms that returned restults
```
crowbot: little typo

crowbot (approved):
Looks good to me!

ajparsons:
If I'm reading it right, there is an optional `min_count` parameter, but doesn't specify a floor if not specified? We might want to set that to 5ish to avoid identifying searches.

PR #124, opened 2021-09-13
data update tweaks

struan:
Slightly better output when deleting a council, plus a script to make it easier to get a list of upcoming DB changes.


crowbot (approved):
Looks good!

PR #125, opened 2021-09-16
Display council powers on council detail page

zarino:
Displays powers for all local authorities (but only very generic output for Combined Authorities, since they’re super varied):

Screenshot 2021-09-16 at 10 15 53

Clicking a power unfolds examples of climate actions that councils might take:

Screenshot 2021-09-16 at 10 16 11

Uses `<details>` elements, so works without JavaScript. If JavaScript is enabled, though, it’s upgraded to work like an accordion, so only one section will be visible at any one time.

Conditionals in the `includes/council-powers.html` template make sure we’re only showing the relevant powers for each type of local authority (eg: no Transport Planning for Non-Metropolitan Districts inside a County; no Social Housing for Counties).

For English councils, we provide a link to read more in the UK100 Power Shift report.


zarino:
The headings are currently in largely random order. I wonder whether they should be alphabetical? Or in order of "most powerful" to "least powerful"?

struan:
This looks lovely.

zarino:
@struan I’ve added some amendments based on CEUK’s feedback in the Google Doc. Think the next step is to get this merged and deployed, then CEUK (and anyone else!) can have a browse with it in-situ and make any remaining comments.

caps/templates/includes/council-powers.html
```diff
@@ -0,0 +1,135 @@
...
+
+{% if council.authority_type != 'COMB' %}
+    <details class="details details-accordion">
+      {% if council.authority_type != 'CTY' and council.authority_type != 'NMD' %}
+        <summary class="mb-2">Waste collection and disposal</summary>
+      {% elif council.authority_type != 'CTY' %}
+        <summary class="mb-2">Waste collection</summary>
+      {% elif council.authority_type != 'NMD' %}
+        <summary class="mb-2">Waste disposal</summary>
+      {% endif %}
```
struan, line 101: This set of conditions is a bit confusing. If I'm correct this might be better expressed as
if not CTY and not NMD
    Waste collection and disposal
elif CTY
    Waste disposal
else
    Waste collection

caps/templates/includes/council-powers.html
```diff
@@ -0,0 +1,135 @@
...
+            <p>Climate actions might include:</p>
+            <ul>
+                <li>reducing the carbon footprint of civic buildings through better insulation and renewable energy use</li>
+                <li>incentivising ‘Active Travel’ or public transport use among employees</li>
+            </ul>
+        </div>
+    </details>
+{% endif %}
+
+{% if council.authority_type != 'COMB' and council.authority_type != 'CTY' %}
```
struan, line 79: Could combine these two ifs

struan (changes requested):
Few small things with the if statements that could do with tidying/clarification.

caps/templates/includes/council-powers.html
```diff
@@ -0,0 +1,135 @@
...
+            <p>Climate actions might include:</p>
+            <ul>
+                <li>prioritising energy efficiency when building new council-operated social housing schemes – for example, through the Passivhaus standard</li>
+                <li>enabling housing associations to improve the energy efficiency of their housing stock, through loans from the local authority</li>
+            </ul>
+        </div>
+    </details>
+{% endif %}
+
+{% if council.authority_type != 'COMB' and council.authority_type != 'CTY' %}
```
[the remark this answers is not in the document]
zarino: Figured I’d leave them separate, as wasn‘t sure what order we wanted the sections in.

caps/templates/includes/council-powers.html
```diff
@@ -0,0 +1,135 @@
+{% if council.authority_type == 'CC' %}
+    <p class="mt-n1">{{ council.name }} has powers over:</p>
+{% else %}
+    <p class="mt-n1">{{ council.name }} is a <strong>{{ council.get_authority_type_display }}</strong>, with powers over:</p>
```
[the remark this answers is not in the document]
zarino: AIUI Scottish and Welsh councils are unitary authorities too, so I think this wording is fine for those.

But you’re right, Northern Irish councils _are_ different. They seem to be described as a _special kind_ of Unitary Authority (for example, by the Institute for Government). I‘ve added a commit that calls them a "Northern Irish Unitary Authority", and prevents us listing the reponsibilities they don’t actually have (transport, schools and libraries, social housing).

caps/templates/includes/council-powers.html
zarino: Done! (Except now it’s done in the model instead of the template, yay!)

caps/templates/includes/council-powers.html
zarino: I’ve restructured the template to do less fancy logic, so this isn’t an issue any more.

zarino:
Thanks for the feedback @struan. I’ve done another tidy. Feel free to merge and deploy if you’re happy.

caps/templates/includes/council-powers.html
```diff
@@ -0,0 +1,154 @@
...
+        <div class="council-powers">
+            <p>Climate actions might include:</p>
+            <ul>
+                <li>making council offices more energy efficient</li>
+                <li>incentivising ‘Active Travel’ or public transport use among employees</li>
+                <li>providing carbon literacy training for employees</li>
+            <ul>
+        </div>
+    </details>
+{% endif %}
```
struan, line 21: given this is always present we could skip the if, similarly with spending below?

caps/models.py
```diff
@@ -68,6 +68,39 @@ def __str__(self):
...
+    def powers(self):
+        powers = ['staff', 'spending']
+
+        if self.country != self.NORTHERN_IRELAND and self.authority_type != 'COMB' and self.authority_type != 'NMD':
+            powers.append('transport-planning')
+
+        if self.country != self.NORTHERN_IRELAND and self.authority_type == 'UA' or self.authority_type == 'MD' or self.authority_type == 'CTY':
+            powers.append('passenger-transport')
+
+        if self.country != self.NORTHERN_IRELAND and self.authority_type != 'COMB' and self.authority_type != 'NMD':
```
struan, line 81: Is there a reason not to merge all the if statements that are the same?

struan: I wonder if this might make more sense to do like
```
mapping = {
    'transport-planning': ['UA', 'CTY', 'CC', 'LBO']
    'passenger-transport: ['UA', 'MD', 'CTY']
...
}

for power, types in mapping.items():
    if self.authority_type in types:
        powers.append(power)

```

and then have a separate if for NI with the list for that.

struan (approved):
some suggestions but I don't think they're actually necessary and the loop thing is more my perl programming "make everything a hash!" history showing :)

caps/models.py
zarino: I thought about a map structure like you suggested, but was a bit worried we might soon want conditions that are more complex than just "council is one of these types". Felt like premature optimisation, given we’re not even sure this is how we want to store powers in the first place?

caps/templates/includes/council-powers.html
zarino: I like the consistency that every power has a conditional, even if it _does_ always evaluate to true in these two cases. Feels less surprising to look at the template and see everything nested in a check, rather than having two of them sit there without a condition.

zarino:
@struan could you rebase, merge, and deploy this when you get a chance? Thanks!

PR #126, opened 2021-09-16
Fix bugs with various last update fields on plan import

struan:
At the moment we're always updating all the plans each time we import even if nothing has changed so this only updates the changed plans, which means the `updated_at` field will be accurate. It also makes sure the first and last found fields are updated in a sensible way.


crowbot (approved):
Looks good to me!

PR #131, opened 2021-09-22
Add climate emergency declarations

struan:
Display if a council has declared a climate emergency in the council list, and on the council detail page.


crowbot (approved):
Looks good to me, thanks!

PR #132, opened 2021-09-23
Small tweaks for future proofing

struan:
Two small changes extracted from the promises branch because they're likely to be useful elsewhere.

1 - move the location of the council page filters to make it easier to expand them.
2 - split out some of the plans CSV processing code into a utils file so it can be used in other import processes.

### DIFF
--- a/caps/management/commands/add_authority_codes.py
+++ b/caps/management/commands/add_authority_codes.py
@@ -8,81 +8,9 @@ import pandas as pd
 from django.core.management.base import BaseCommand, CommandError
 from django.conf import settings
 
-AUTHORITY_MAPPING_URL = 'https://raw.githubusercontent.com/crowbot/uk_local_authority_names_and_codes/master/lookup_name_to_registry.csv'
-AUTHORITY_MAPPING_NAME = 'lookup_name_to_registry.csv'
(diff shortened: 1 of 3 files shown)


caps/management/commands/add_authority_codes.py
```diff
@@ -108,39 +36,17 @@ def add_missing_authorities():
...
         print('getting data files')
         get_data_files()
         print('adding authority codes')
-        add_authority_codes()
+        add_authority_codes(settings.PROCESSED_CSV)
         print('adding authority info')
-        add_extra_authority_info()
+        add_extra_authority_info(settings.PROCESSED_CSV)
         print('adding missing authorities')
         add_missing_authorities()
```
crowbot, line 50: Should this also be passed `settings.PROCESSED_CSV` ?

crowbot (approved):
Looks good to me, one comment inline

caps/management/commands/add_authority_codes.py
struan: I wasn't sure that the was really needed for anything other than the plans sheet so I didn't include it in the changes.

zarino:
Nice to see that rejigged council page commit finally work its way out into the wild!

struan:
Merged as of e963d96

PR #133, opened 2021-10-01
Internal mySociety deployment integration

sagepe:
This makes some changes to the project to better integrate with our internal systems management and deployment automation.

A side effect is the switch away from having two separate compose files in the root of the project, so I've also renamed the development file to `docker-compose.yml` as this conforms with the general standard naming convention.


zarino (approved):
Nice no longer having to remember `-f development.yml` all the time! 👍 

All worked with a `docker-compose up` after checking out this branch. I obviously already had the containers set up, from the docker config in `master`.

struan (approved):

PR #136, opened 2021-10-04
Limit search by council name

### LINKED ISSUE #128: Limit search by council, and campaign pre-population
Currently the text search is across the database and we don't have modifiers to restrict to a single council or set of councils. 

[The set of councils issue can await wider thinking of how we are managing filters and comparisons]. 

The potential use cases:

1. Provide a search box on a council page to let you that search the council's documents without downloading. 
2. Provide a campaign pre-population view, where I could create a url to distribute to "find out what your council's plan says about household insulation". This would have two text boxes, one prepopulated with the search term, and the other a postcode field. [The logical extension of this is a 'share this search' button on the results page, that lets you produce a campaign].

struan:
Adds a text box to allow a user to limit the text search to councils matching the text.

Also saves this restriction field when saving search terms.

Fixes #128.


struan:
@sagepe It looks like there's not a reasonable way to get docker to rebuild if you change the schema, except maybe by having our own custom image which seems maybe like overkill (?) so instead I've written a script that needs to be run by hand to update the schema and tell docker to reload the core. Given that I don't see schema updates being a frequent thing this seems reasonable?

struan:
@zarino this is now at a stage where it could do with some design attention.

caps/templates/search_results.html
```diff
@@ -2,38 +2,74 @@
...
-  {% if page_obj.has_previous or page_obj.has_next %}
-    <div class="d-flex justify-content-center align-items-center mt-5">
-      {% if page_obj.has_previous %}
-        <a class="btn btn-outline-primary mx-2" href="?q={{ query }}&amp;page={{ page_obj.previous_page_number }}">
-            Previous results
-        </a>
+      {% if page_obj.has_previous or page_obj.has_next %}
+        <div class="d-flex justify-content-center align-items-center mt-5">
+          {% if page_obj.has_previous %}
+            <a class="btn btn-outline-primary mx-2" href="?q={{ query }}&amp;page={{ page_obj.previous_page_number }}">
```
zarino, line 57: @struan should these next/previous buttons also contain a query param for `council_name`?

docker-compose.yml
```diff
@@ -38,6 +38,7 @@ services:
       - solr_data:/opt/solr/server/solr/mycores
       - ./conf/schema.xml:/opt/solr/server/solr/configsets/basic_configs/conf/schema.xml
       - ./conf/solrconfig.xml:/opt/solr/server/solr/configsets/basic_configs/conf/solrconfig.xml
+      - ./script/update_solr_schema:/opt/docker-solr/scripts/update_solr_schema
```
zarino, line 41: Should we document how/when to run this, somewhere? Like in the Readme?

zarino (commented):
I‘ve added a commit that styles up the results page. Now there’s a second text input, it made sense to follow through with the same "form in the sidebar" layout that we use on the council list page.

Screenshot 2021-11-19 at 11-33-15 swimming – Search results

Couple of minor comments inline.

zarino (approved):
:+1: LGTM!

PR #139, opened 2021-10-19
fix missing council urls

struan:
either by updating the alternative names or in the case of NI just hard
coding them


crowbot (approved):
Looks great, thanks!

PR #140, opened 2021-10-21
Update emissions data to 2019 version

struan:
Adds a replace option as the new data uses slightly different headings so it makes more sense to replace everything.

### DIFF
--- a/caps/dataframe/la.py
+++ b/caps/dataframe/la.py
@@ -1,0 +1,238 @@
+"""
+functions and pandas api to speed up working with
+local authority data
+
+"""
+from functools import lru_cache
+from typing import Callable, Optional, List
+import string
+import pandas as pd
+import numpy as np
+
+la_lookup_url = "https://raw.githubusercontent.com/ajparsons/uk_local_authority_names_and_codes/master/uk_local_authorities.csv"
+name_lookup_url = "https://raw.githubusercontent.com/ajparsons/uk_local_authority_names_and_codes/master/lookup_name_to_registry.csv"
+gss_code = "https://raw.githubusercontent.com/ajparsons/uk_local_authority_names_and_codes/master/lookup_gss_to_registry.csv"
+
+
+@lru_cache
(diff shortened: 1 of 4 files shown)


crowbot (changes requested):
Running this under python 3.7 with `python3 manage.py import_emissions_data --replace`, I get an error: 
```
 File "/vagrant/caps/caps/management/commands/import_emissions_data.py", line 14, in <module>
    import caps.dataframe.la
  File "/vagrant/caps/caps/dataframe/la.py", line 18, in <module>
    def get_la_df(include_historical: bool = False) -> pd.DataFrame:
  File "/usr/lib/python3.7/functools.py", line 477, in lru_cache
    raise TypeError('Expected maxsize to be an integer or None')
TypeError: Expected maxsize to be an integer or None
``` 

which it [looks like](https://stackoverflow.com/questions/47218313/use-functools-lru-cache-without-specifying-maxsize-parameter) can be fixed by adding parenthesis to the decorator calls to `lru_cache`. Then, though, I get a warning: 

``` 
/vagrant/caps/caps/management/commands/import_emissions_data.py:142: FutureWarning: Sorting because non-concatenation axis is not aligned. A future version
of pandas will change to not sort by default.

To accept the future behavior, pass 'sort=False'.

To retain the current behavior and silence the warning, pass 'sort=True'.

  return pd.concat([df, county_df, combined_df]).drop(columns=["Year"])
```
Would be good to resolve what our desired behaviour is here before merging.

ajparsons:
It doesn't really matter because the sorting here is about column order, and everything is accessed by column name in follow up steps. So the new default behaviour of `sort=False` is fine. I can add that, but given the pandas version in the requirements.txt has that behaviour, it seems ok to me?

struan:
With the django 3 upgrade it also updated the version of pandas to the latest, along with python to 3.9, so is this possibly a version thing?

struan:
merged in a9423636de6fe9d2272ba348cd283632170489e9

PR #142, opened 2021-10-21
Display and sort by last plan update date on council list page

struan:


struan:
And now we've added more data in this definitely needs a rethink:

Screenshot 2021-11-02 at 15 41 09

zarino:
@struan On the `council_list` page, do you think we could hide the "last updated" column by default, and only show it if the user has chosen to sort the table by "Last update" or "Last update (descending)"? If that’s possible, we could hide the carbon neutral target and emergency declaration columns to make room, and switch them back when they sort by anything else.

I’ve added a commit which moves the "Nearby councils" table into the primary page column, giving it more room. (Technically part of #157.) Ironically, don’t know how useful it is to show the last updated date on this table, so I’ve left it hidden. But at least the columns will lay out more nicely now on mid-sized screens.

struan:
@zarino done that. it's does seem a bit against the principle of least surprise.

zarino:
@struan Yeah, it’s a little bit weird. The way I’m justifying to myself is that "last updated" is a bit of a power-user field, since it’s referring to _our process_ rather than anything inherent to the council. The time it’ll come in useful is when you’re looking for the most or least recently updated councils – so we reveal it when you’re sorting by date.

Can always reconsider when we inevitably have _another_ column to add to this table! 😉

zarino (approved):
Love it! :shipit:

PR #151, opened 2021-11-02
Track clicks on Net Zero Local Hero landing page

zarino:
More tracking on the Net Zero Local Hero landing page.

Records a Google Analytics event whenever any `a[href]` on the landing page is clicked (_anywhere_ on the page, including the header and nav). Sends Google the text content of the link itself (or the image filename, if it’s a link containing only an image) and the URL that the user is about to go to.

Doesn’t track form submissions (like the postcode search form) because that’s a bit harder.

Not very pretty code, but good enough to temporarily measure activity on the landing page over COP26.

/cc @MyfanwyNixon

### DIFF
--- a/caps/static/js/main.js
+++ b/caps/static/js/main.js
@@ -272,3 +272,31 @@ var trackEvent = function(eventName, params) {
 $('.details-accordion').on('click', function(){
     $(this).siblings('.details-accordion[open]').removeAttr('open');
 });
+
+$('.nzlh-landing-page').on('click', 'a[href]', function(e){
+    var href = $(e.currentTarget).attr('href');
+
+    var callback;
+    if (
+        e.ctrlKey ||
+        e.shiftKey ||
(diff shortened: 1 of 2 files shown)


caps/static/js/main.js
```diff
@@ -272,3 +272,31 @@ var trackEvent = function(eventName, params) {
 $('.details-accordion').on('click', function(){
     $(this).siblings('.details-accordion[open]').removeAttr('open');
 });
+
+$('.nlzh-landing-page').on('click', 'a[href]', function(e){
```
struan, line 276: `.nzlh-landing-page` surely?

struan (approved):
one small thing but this looks good.

caps/static/js/main.js
zarino: God dammit. I never get these letters the right way round. Good spot!

PR #153, opened 2021-11-04
move information about data to its own page

struan:
Mostly so it doesn't overwhelm the about page and also so as we expand
the data we can put more information here.


zarino (approved):
This looks good!

Issue #145, opened 2021-10-25
think about how to archive old plans


struan:

zarino:
@struan I see https://github.com/mysociety/caps/pull/159 was merged, but remind me, this still isn’t "done" right? ie: we don’t store old versions of plans, and they’re not accessible somehow?

struan:
We never delete the files so they are still on the disc, but there's no way to access them, no.

I have a branch somewhere that has a proof of concept of how to display this which I can tidy up and push.

struan:
I've updated the PR so it will now link to deleted files that we still have on disk, although it possibly needs an extra indication to show it's an old plan that is no longer current.

JenMysoc:
This is in For Review - but I'm not sure if it's been reviewed and deployed?

zarino:
@JenMysoc Yeah, it’s lain dormant for a while. I mentioned to @lucascumsille at Design sprint planning yesterday, I’ll check this out this sprint, and either fix it up and merge, or just merge as-is.

PR #159, opened 2021-11-09
add django-simple-history to save plan changes

struan:
Only tracks changes to plans and promises at the moment, doesn't do anything with
them.

Note, should run `manage.py populate_history --auto` on initial deploy so there's a base to compare changes to.

Part of #145


crowbot (commented):
Seems to run OK - doesn't look like we're exposing the history anywhere though?

struan:
It's not exposed anywhere at the moment. I wanted to at least be tracking the changes so we don't lose them and we can sort out how to use it later.

PR #165, opened 2021-11-16
basic council climate league app

struan:
This uses django hosts so it required a rebuild of the docker containers (`docker compose build`) and then if you use a host with 'league' in it (e.g `http://league.127.0.0.1.nip.io:8000`) then it will use the league app. This includes a basic index view plus the SCSS file and base template.


struan:
@sagepe I would like to work out a way to test this as deployed on our servers

PR #163, opened 2021-11-15
Council detail page layout refresh

zarino:
* Move council name out of the sidebar, into a full-width header, fixing #113
* Move the council homepage link into the ‘More about this council’ box, next to all the other external links
* Swap the DOM order of the sidebar and main content, improving mobile UX and fixing the primary issue in #157
* Combine the ‘Climate emergency’ and ‘Headline promises’ boxes into a single ‘Declarations & target dates’ box, to save vertical space
* PDF icons for PDFs!

Screenshot 2021-11-15 at 16-24-25 East Cambridgeshire District Council

Screenshot 2021-11-15 at 16-24-19 Newcastle Upon Tyne, North Tyneside and Northumberland Combined Authority

PR #166, opened 2021-11-16
Make sure even “other” scope promises have a heading

zarino:
Fixes a bug introduced in #163 that caused promises (“pledges”) with a scope of “other” to display on the council detail page without a heading above them, meaning they crashed into the line above.

**Before:**

Screenshot 2021-11-16 at 14 24 00

**After:**

Screenshot 2021-11-16 at 14 23 41


struan (approved):
👍

PR #170, opened 2021-11-23
Add an exact search option to the search form

### LINKED ISSUE #168: Unexpected Solr stemming results in, eg: "expenses" (in quotes) returning results for "expensive"
There’s probably nothing we can do about overzealous stemming for queries like `expenses` returning results for `expens*`, but I think most people would expect, when they put double quotes around their search term, it should return only _exact_ phrase matches.

Yet, a search for `"expenses"` (https://data.climateemergency.uk/search/?q=%22expenses%22) still includes results for words other than `expenses`.

Is this something we can tweak in the Solr setup?

struan:
This required an extra solr field with unstemmed text which will increase the size of the index but as it's never going to get that large I think this is a reasonable trade off. You can't just do an Exact on a stemmed field as it still stems.

Fixes #168

### DIFF
--- a/caps/forms.py
+++ b/caps/forms.py
@@ -1,8 +1,11 @@
-from django.forms import CharField
+from django.forms import CharField, BooleanField
 from haystack.forms import SearchForm
+from haystack.query import SearchQuerySet
+from haystack.inputs import Exact
 
 class HighlightedSearchForm(SearchForm):
     council_name=CharField(required=False)
+    exact=BooleanField(required=False)
 
     def search(self):
         kwargs = {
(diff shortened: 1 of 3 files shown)


zarino:
Oh, so I take it from this, that Solr just won’t support exact search matches via double quotes in the search string?

struan:
It will not, or at least I could not persuade it to do so.

zarino (approved):
A shame this requires a whole separate checkbox, but hey ho. I’ve added a commit that hopefully makes it clearer what this option does (let me know if I’ve misrepresented it!) and also tidies up some of the visual hierarchy in that form now there’s more than one optional input.

Screenshot 2021-11-24 at 18 08 23

crowbot (changes requested):
Thanks both. This looks good - although @zarino the advanced search layout seems broken for me in Chrome - as below (was fine in FF)
Screenshot 2021-11-25 at 09 12 01

I think it could be better not to use 'stemming' in an interface - it's a bit jargon-y. Maybe we could lose the heading altogether and just have the text say "By default, search results include other words that are similar to your search phrase. If you want, you can search for your exact phrase only"

I feel like it _ought_ to be possible to get Solr/Lucene to just do the unsurprising thing with double quotes, but perhaps that's just dreaming the impossible dream. This is a good improvement.

In reviewing this, I noticed that the LGA have removed (temporarily, perhaps?) their page on combined authorities, ticketed [here](https://github.com/mysociety/caps/issues/172).

zarino:
@crowbot I’ve fixed the Chrome `legend` spacing bug (turns out vertical margins behave weirdly on `legends` in Chrome under certain circumstances, and Bootstrap doesn’t normalise this) and changed the "Search stemming" title to "Similar words".

zarino:
@crowbot Odd, I wasn‘t seeing that, unless I disabled JavaScript. Still, wasn’t intentional! Have pushed another fixup, and have tested in all the modern browsers and IE11 too.

struan:
Merged at 3a85106fadad9a1ea935e2b719a5fb1cddb741fa

PR #195, opened 2021-12-09
Adjust config for codespaces

ajparsons:
Generally this just changes the vscode config files, but there are two changes to the caps setup process in the latest commits.

These fix a future deprication warning for pandas, and add an explicit user agent for wikidata. This is technically [required anyway](https://meta.wikimedia.org/wiki/User-Agent_policy) but we've been getting away with it given our low traffic IPs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,36 @@ You can rebuild the application container by running `script/build`. Bear in min
 
 The environment will be visible at http://localhost:8000 and the Solr admin interface at http://localhost:8983
 
+### Codespaces/VSCode
+
+In your [Github Codespaces settings](https://github.com/settings/codespaces), set up the relevant secrets (mostly MAPIT_API_KEY) and give the caps repo access to this secret.
+
+Then in the [caps repo](https://github.com/mysociety/caps/), click the code dropdown (top right), and select the codespaces tab, then create new codespace. 
+
+Once it has cloned and set up the docker configuration (it may prompt you to reload with the pylance extension, this is fine and takea a few seconds). You have two options:
+
+To run the data population from scratch run `script/setup-in-docker` (as warned above, this takes a while).
+To restore a previous database dump, run `script/quick-setup`.
+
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -44,6 +44,36 @@ You can rebuild the application container by running `script/build`. Bear in min
...
+
+The development docker image used by vscode/codespaces is based on a prebuilt version of the main Dockerfile to save start-up time.
+
+This means some adjustments are necessary when a change to the branch involves a change to the base image (e.g. new packages or changed version). 
+
+In this event, commit the changes that affect the image to the branch, and go to [the workflow page](https://github.com/mysociety/caps/actions/workflows/main.yml) that can create a new image. 
+
+In this page click run workflow, and use the workflow from your new branch. 
+
+This will build the docker image and tag it with the name of the your branch. This will appear in the [package page](https://github.com/mysociety/caps/pkgs/container/caps) when ready. If the build fails, the details are in the actions tab.
```
struan, line 70: How does this work with people outside mySociety forking this repo?

struan: Not that it's an issue at the moment, mostly just curious.

struan (approved):
One question but I think it might be irrelevant?

README.md
ajparsons: Hmm. My understanding of this is the workflow action would go with the forked repo, if they ran it it would work fine and generate a new caps image in their forked repos packages. What they'd then need to do is a slightly different amend to `dockerfile.dev` and change the 'mysociety' bit in `ghcr.io/mysociety/caps:master` to their name.

I guess the tricky bit would be if we then merged back in a pull request, we'd want to make sure we weren't changing our image chain to be based on something now outside our control. I guess an approach to this would be a github action that checks if it's still the `mysociety` version of caps, and then checks the first line refers of that repo points to the right place. If it doesn't, test and pull request fails. But think I'll choose not to worry about that at this point...

PR #197, opened 2021-12-09
Changed datasources to point to new repo

ajparsons:
We now have a [repo](https://github.com/mysociety/uk_local_authority_names_and_codes) with a good source of truth for local authority information including combined authorities. 

This PR removes and updates references to older versions of this lookup, and added information that was being manually patched back into the main repo. 

This seems to setup from scratch ok, but @struan might have some other ideas about edge cases that are worth testing.

### DIFF
--- a/caps/import_utils.py
+++ b/caps/import_utils.py
@@ -12,28 +12,26 @@ import ssl
 ssl._create_default_https_context = ssl._create_unverified_context
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
 
-AUTHORITY_MAPPING_URL = 'https://raw.githubusercontent.com/crowbot/uk_local_authority_names_and_codes/master/lookup_name_to_registry.csv'
+
+AUTHORITY_MAPPING_URL = 'https://raw.githubusercontent.com/mysociety/uk_local_authority_names_and_codes/main/data/lookup_name_to_registry.csv'
 AUTHORITY_MAPPING_NAME = 'lookup_name_to_registry.csv'
 AUTHORITY_MAPPING = join(settings.DATA_DIR, AUTHORITY_MAPPING_NAME)
 
-AUTHORITY_DATA_URL = 'https://raw.githubusercontent.com/crowbot/uk_local_authority_names_and_codes/master/uk_local_authorities.csv'
+AUTHORITY_DATA_URL = 'https://raw.githubusercontent.com/mysociety/uk_local_authority_names_and_codes/main/data/uk_local_authorities.csv'
 AUTHORITY_DATA_NAME = 'uk_local_authorities.csv'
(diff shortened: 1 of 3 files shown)


caps/import_utils.py
```diff
@@ -51,13 +49,15 @@ def add_authority_codes(filename):
     for index, row in plans_df.iterrows():
         council = row['council'].lower()
 
-        name_versions = [council, council.replace('council', '').strip(), council.replace('- unitary', '').replace('(unitary)', '').strip()]
+        name_versions = [council, council.replace('council', '').strip(
+        ), council.replace('- unitary', '').replace('(unitary)', '').strip()]
```
struan, line 53: one small request is that in future could formatting only changes and code changes be two separate commits?

Ideally I guess we should just take a one off formatting hit across the whole repo and then agree on formatting rules that we apply.

caps/import_utils.py
```diff
@@ -69,59 +69,39 @@ def add_gss_codes(filename):
...
+    df = df.rename(columns={"local-authority-code": "authority_code",
+                            "local-authority-type": "authority_type",
+                            "wdtk-id": "wdtk_id",
+                            "mapit-area-code": "mapit_area_code",
+                            "nation": "country",
+                            "gss-code": "gss_code"
+                            })
 
-    aliases = {'newcastle upon tyne, north tyneside and northumberland combined authority': 'north of tyne'}
+    df = plans_df.merge(df, how="left")
```
struan, line 96: A comment to indicate the basis of this merge would be helpful. I assume it's the gss_code? Or does it just need to have one of the columns to match?

struan (approved):
Much less code so yay!

One small comment request and a request/thought on future formatting.

caps/import_utils.py
ajparsons: Sorry I missed the autoformatter doing this. My defaults seems to be to use autopep8 with a few rules ignored. Could have a chat next week about if that's a good approach for a one off fix or if there's other rules to ignore?

caps/import_utils.py
ajparsons: Sure, it's the intersection of the columns of the two dataframes (so in this case 'authority_code'). I'll change this so the merge on is explicit.

ajparsons: Addressing this also led to a change to explicitly drop the columns that are going to be merged in (as the csv may have been updated previously, and may already contain them).

caps/import_utils.py
struan: noiseworks look to be using black which seems like it gets round a lot of the "what rules do we use?" questions. I don't really mind what we use, as long as everyone is doing the same thing so we avoid reformatting wars :)

ajparsons: Ooo, that looks quite nice.

struan: Yes, I think that would work.

PR #199, opened 2021-12-14
Scoring mvp

struan:
This covers four things:

    import of scores
    front page with total and section scores
    all questions results
    all answers for a question

once you've checked this out then you can do the usual django migrate.

You'll then need two extra files to go in the data directory:

    council_plan_scores.csv
    all_councils_scored.csv
    Probably easiest to ask me for a copy of each of these until they are somewhere public to download.

you can then run manage.py import_plan_scores and that will set up the data. Do not worry if it has a lot of messages about not finding scores.


struan:
no longer required

PR #203, opened 2021-12-16
Import plan scores

struan:
At the moment this doesn't download two of the files required because there's not a place for them so you'll need those to test the files. You can get those here: https://drive.google.com/drive/folders/1Zjwfsa2r5_v-0HsgG0WiPxJaDmlhEId4


scoring/models.py
```diff
@@ -1,3 +1,56 @@
...
+    plan_section = models.ForeignKey(PlanSection, on_delete=models.CASCADE)
+    score = models.PositiveSmallIntegerField(default=0)
+    weighted_score = models.PositiveSmallIntegerField(default=0)
+
+    @classmethod
+    def get_all_council_scores(cls):
+        scores = cls.objects.all().select_related('plan_section', 'plan_score').filter(plan_score__total__gt=0).values('plan_score__total', 'plan_score__council_id', 'score', 'weighted_score', 'plan_section__code', 'plan_section__max_score')
+        councils = defaultdict(dict)
+        for score in scores:
+            if score['plan_section__code'].find('(') != -1:
```
crowbot, line 38: What's this handling? As I read it, it's looking for PlanSections with a code that has a `(` in it - how would that come about?

scoring/management/commands/import_plan_scores.py
```diff
@@ -0,0 +1,174 @@
...
+            except PlanSection.DoesNotExist:
+                print('no section found for q {}, section {}'.format(code, section))
+                continue
+
+            question, created = PlanQuestion.objects.get_or_create(
+                code=code,
+                section=plan_section
+            )
+            if created:
+                print(code)
```
crowbot, line 124: Do you want this debug code left in?

scoring/management/commands/import_plan_scores.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def import_section_scores(self):
+        council_scores = {}
+
+        df = pd.read_csv(self.SCORES_CSV)
+        for index, row in df.iterrows():
+            if row['section'] == 'overall':
+                continue
+
+            weight = PlanDocument.integer_from_text(row['weighting'])
```
crowbot, line 47: We should really move these "x_from_text" functions to a utility module now they're not specific to `PlanDocument`

scoring/management/commands/import_plan_scores.py
```diff
@@ -0,0 +1,174 @@
...
+            max_score = PlanDocument.integer_from_text(row['max_score'])
+
+            section, created = PlanSection.objects.get_or_create(
```
crowbot, line 50: Worth a comment saying what's happened here? i.e. that we're creating the section on-the-fly if we haven't already got it?

scoring/management/commands/import_plan_scores.py
```diff
@@ -0,0 +1,174 @@
...
+                section.description = row['section_name']
+                section.max_score = max_score
+                section.max_weighted_score = max_score * weight
+                section.weight = weight
+                section.save()
+
+            try:
+                council = Council.objects.get(authority_code=row['council_id'])
+            except Council.DoesNotExist:
+                continue
```
crowbot, line 64: Feels like we might want some output here to say that we expected a council that we didn't find?

scoring/management/commands/import_plan_scores.py
```diff
@@ -0,0 +1,174 @@
...
+
+            plan, created = PlanScore.objects.get_or_create(
```
crowbot, line 66: Can we call this `plan_score` rather than `plan`, to avoid confusion?

scoring/management/commands/import_plan_scores.py
```diff
@@ -0,0 +1,174 @@
...
+                code=code,
+                section=plan_section
+            )
+            if created:
+                print(code)
+                max_score = 0
+                if q_type != 'HEADER':
+                    scores = PlanDocument.char_from_text(row['Scores'])
+                    scores = scores.split(',')
+                    print(scores)
```
crowbot, line 129: Debug code

scoring/management/commands/import_plan_scores.py
```diff
@@ -0,0 +1,174 @@
...
+                    max_score = max(scores)
+                print(max_score)
```
crowbot, line 131: Debug code

proj/settings.py
```diff
@@ -152,6 +152,16 @@
 DECLARATIONS_CSV_NAME = 'declarations.csv'
 DECLARATIONS_CSV = os.path.join(DATA_DIR, DECLARATIONS_CSV_NAME)
 
+PLAN_YEAR = 2021
+PLAN_SCORE_CSV_KEY = ''
+PLAN_SCORE_QUESTIONS_CSV_KEY = '1NkAcDZ0wZwnb7iiwWSy5gMf7rJfhDPnDWvZtaJ4l0f8'
+PLAN_SCORE_QUESTIONS_CSV_TAB = 1790236873 # this number obtained from browser network console :|
+PLAN_SCORE_ANSWERS_CSV_KEY = ''
+PLAN_SCORE_CSV_NAME = 'council_plan_scores.csv'
```
crowbot, line 160: Could be worth renaming this to convey that it's the section scores

scoring/management/commands/import_plan_scores.py
```diff
@@ -0,0 +1,174 @@
...
+
+            try:
+                council = Council.objects.get(authority_code=council_code)
+                plan = PlanScore.objects.get(council=council, year=self.YEAR)
+                question = PlanQuestion.objects.get(code=code)
+            except Council.DoesNotExist as e:
+                print('failed to match council {}'.format(council_code))
+                continue
+            except PlanScore.DoesNotExist as e:
+                print('failed to match plan {}'.format(council.name))
```
crowbot, line 153: Think this error should be `failed to match plan score for {}`

scoring/management/commands/import_plan_scores.py
```diff
@@ -0,0 +1,174 @@
...
+    def import_question_scores(self):
+        df = pd.read_csv(self.ANSWERS_CSV)
+        for index, row in df.iterrows():
+            code = self.normalise_section_code(row['question_id'])
+            council_code = row['answer_id']
+            council_code = re.sub(r'^([^_]*)_.*', r'\1', council_code)
+
+            try:
+                council = Council.objects.get(authority_code=council_code)
+                plan = PlanScore.objects.get(council=council, year=self.YEAR)
```
crowbot, line 147: Would be good to call this `plan_score` rather than `plan`

crowbot (changes requested):
Looks good overall - I can't test `import_question_scores` yet as the test directory only has a section score file and a question details file, not a question answers file. 

I wrote the following cheatsheet for myself to understand what the models are:
```
PlanQuestion - details of a question in the scoring
PlanScore - overall score for a council’s plan
PlanSection - details of a section in the scoring
PlanSectionScore - score for a plan for a section
PlanQuestionScore - score for a plan for a question 
```
If that's right, might be worth adding as descriptions to the classes in `models.py` so it's there for easy reference later

scoring/models.py
struan: doh, that was dealing with a dumb thing I did and should have been removed.

scoring/management/commands/import_plan_scores.py
struan: I've created an issue for this rather than do it here: https://github.com/mysociety/caps/issues/214

struan:
You should be able to test with just those two as it should download the question answers file.

crowbot (commented):
Fixups look good - I think I just need the answer file to be able to check that last function

crowbot (approved):

Issue #176, opened 2021-12-02
[CCL] Header, footer, and base styles and markup


zarino:
- [x] Use Bootstrap
- [x] Nav bar with menu items and a search box
- [x] Footer with CE UK branding (@lucascumsille)

### Related

#177 – Header search box

lucascumsille:
- [ ] Change font-size H3 on mobile to 18px (@lucascumsille)
- [ ]  Change font-size body to 13px-15px (@lucascumsille)

PR #210, opened 2021-12-17
Added basic boostrap styling and navbar

lucascumsille:
Fixes part #176


scoring/static/scss/scoring.scss
```diff
@@ -57,4 +57,22 @@
 @import "../bootstrap/utilities";
 @import "../bootstrap/print";
 
-@import "mysociety-standard-footer";
+//The variables.scss is not being imported so I'm putting some variables here.
```
zarino, line 60: @struan can you help @lucascumsille work out what’s going wrong here with the Sass includes? It‘s either something to do with a conflict between caps/scoring filenames and bootstrap filenames, or a conflict between caps and scoring filenames… or we’re just not using the correct relative path when importing.

zarino (approved):
Overall looks ok. A couple of weird things (oddly specific degrees and percentages in `.bg-grey-gradient`, empty file at `icons.scss`, commented-out code left at the end of `input.scss`) but nothing show-stopping enough to prevent us just merging this in as-is and tidying up later.

scoring/static/scss/scoring.scss
struan: it's importing fine, some of the colours defined in scoring.scss aren't in variables which seem to be the root of the problem. if you add

```
$color-league-green-l1: #E4FAF5;
$color-league-green-d1: #0BC7A0;
$color-league-blue-l2: #D8F2FE;
$color-league-blue-l1: #7BD2F8;
$color-league-blue-d1: #045C88;
```

to variables then everything seems to work fine.

lucascumsille:
I just added the subheader(Donation strip)
Screenshot 2021-12-22 at 07 13 34

zarino (approved):
:+1:

PR #213, opened 2021-12-20
always assume day comes first in dates

struan:
avoids issues with ambiguous dates.


crowbot (approved):
Looks good to me, thanks!

PR #217, opened 2021-12-22
Tests for climate plan scoring

struan:
Some initial tests for the council plan scoring

### DIFF
--- a/scoring/fixtures/test_answers.json
+++ b/scoring/fixtures/test_answers.json
@@ -271,6 +271,17 @@
     }
   },
   {
+    "model": "scoring.planquestion",
+    "pk": 3,
+    "fields": {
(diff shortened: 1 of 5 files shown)


crowbot (approved):
Looks good to me. In working through the logic that these tests apply to, I did notice that `PlanSectionScore.get_all_council_scores` returns in the hash for each council the weighted score for the council in the section (with the `score` key), but the unweighted max score (with the `max` key) - in line 82. Is that what's intended there? 

```
councils[score['plan_score__council_id']][score['plan_section__code']] = { 'score': score['weighted_score'], 'max': score['plan_section__max_score'] }
```

struan:
Since doing this the import data has changed a little bit so I'm going to rework this based on that.

PR #220, opened 2022-01-04
Docker image fix from end of December 2021

zarino:
Not sure whether this is required? But I needed it to get the site set up on my Mac in December 2021. @struan to investigate.


sagepe:
This is fine, we always build the app image locally anyway, so we might as well do this I think.

sagepe (approved):

struan (approved):
I can't see what harm this can do.

Issue #186, opened 2021-12-02
[CCL] Council page: Everything above the tables


zarino:
- [x] Council name, with a little star and wording if they’re one of the "Top" councils of that type (@lucascumsille)
- [x] Statistics / attributes for the council (the demographic/political/geographical values we’re already storing) (@struan)
   - ❓Do we need a way to explain (or link to explanation of) complex terms, like Deprivation Index
- [ ] Any introductory text content required from Myf / CE UK?
- [ ] Direct link(s) to assessed plan(s) – PDFs or webpages (@struan)

PR #230, opened 2022-01-07
Added hero section for council page

lucascumsille:
Fixes part of #186


zarino (approved):
:+1: Looks good to me. I’ve added a commit that actually pulls through the real council name, type, and country. Not sure we have the other data yet (annual income, political control, etc) in an easily-accessible location. Can add it later.

Issue #191, opened 2021-12-02
[CCL] Question page: table of scores for an individual question


zarino:
- A row per council. Columns for question number, question wording, and this council’s score for the question.
- No search, no filters.

PR #231, opened 2022-01-07
Added about section for question page

lucascumsille:
Also fixed hover effect for the about section in council page.

Fixes part of: #186 and #191


zarino (approved):
:+1: The question page is obviously still very much a work in progress, but this gets us a bit closer, so happy to merge as-is.

PR #233, opened 2022-01-07
Minor fixes to styling

lucascumsille:
-Fixed background-color in the single council tables
-Table-header in home page
-Margins on the single council page.


zarino (approved):
🚀

Issue #219, opened 2022-01-04
Search within council plans and similar council plans


ajparsons:
The search query can limit to a specific council or to similar councils to a council, but there is no search option on the council page itself to search within that council or similar councils plans. 

The search page itself does not expose the 'search within similar councils' option. 

This option under similar councils:

![image](https://user-images.githubusercontent.com/8157058/148058662-6849a991-11e1-4d5a-9bfd-2b2164c5752c.png)

Currently doesn't work because the setting is preserved if someone then uses the form on that page to do a search.

ajparsons:
Second part of this addressed - but still not an option to search the council's own plans on the council page.

PR #234, opened 2022-01-07
Search form can search for similar councils

ajparsons:
When following the link from the council page the name is displayed and passed back through. The comparison type is given in a dropdown.

Addresses part of https://github.com/mysociety/caps/issues/219, where the search link for similar types isn't actually usable. 

If not following the link from the council page, these options aren't visible. 

image

### DIFF
--- a/caps/forms.py
+++ b/caps/forms.py
@@ -1,16 +1,27 @@
-from django.forms import CharField, BooleanField
+from django.forms import CharField, BooleanField, ModelChoiceField, Select
 from haystack.forms import SearchForm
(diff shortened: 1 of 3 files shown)


zarino (approved):
@ajparsons I‘ve added a fixup commit that styles the select box, and hides the "council name includes" form field when you’re already searching for similar councils (I figure you probably wouldn’t use both at the same time, and it cuts down on clutter—temporarily—until we get time to give this search form some UX attention).

search-form

If you’re happy with that, feel free to squash, merge, and deploy :-)

PR #237, opened 2022-01-10
191 question page table

lucascumsille:
Screenshot 2022-01-10 at 15 58 42


scoring/static/scoring/scss/mixins.scss
```diff
@@ -133,3 +133,5 @@ $border-priority-3: 1px solid transparentize(#9CBFB7, 0.8);
 @mixin border-priority-3($direction) {
     border-#{$direction}: $border-priority-3;
 }
+
+$border-priority-2-grey:1px solid transparentize($black, 0.8);
```
zarino, line 137: Missing a space after the `:` colon here, and a newline at the end of the file.

…Actually, why is this variable here (in `mixins.scss`) anyway, rather than in `variables.scss`?

zarino (changes requested):
On the council page, the question IDs are displayed like this:

Screenshot 2022-01-13 at 09 28 26

While, on the question page, they’re displayed this like:

Screenshot 2022-01-13 at 09 28 39

We should be consistent. Which one is right?

---

Speaking of which, the "Q1" on this row is quite a small hit box, and is the only way to access the Question page for that question.

Screenshot 2022-01-13 at 09 48 58

What do you think about making the _entire Question wording_ a link to the Question page?

---

The `overflow-y: scroll` on `.question-content` is causing unnecessary scrollbars:

Screenshot 2022-01-13 at 09 38 23

While chaging the `overflow-y: scroll` to `overflow-y: auto` would fix this, it’d still leave people scrolling long question descriptions through a 200px tall letterbox.

My instinct is that we should truncate the descriptions to a manageable length (the first paragraph?) and have either a JavaScript-powered "More…" button to reveal the initially truncated section, or just a "More…" link that leads to the individual Question page for that question, where the full description would be shown.

---

I’m not sure what these "X out of Y" values are meant to be telling me:

Screenshot 2022-01-13 at 09 44 16

I _assume_ it’s "X out of Y councils got full marks on this question", but we don’t explain that anywhere. Maybe we should? …

Screenshot 2022-01-13 at 09 47 04

lucascumsille:
> On the council page, the question IDs are displayed like this:
> 
> Screenshot 2022-01-13 at 09 28 26
> 
> While, on the question page, they’re displayed this like:
> 
> Screenshot 2022-01-13 at 09 28 39
> 
> We should be consistent with the format of the IDs. `Q1` or `S1_GOV_Q1`, which one is right?

In this one I'm not really sure. I took the code that was already on the page. However I prefer the S1_GOV_Q1 because it tells in which section you are in.
> 
> Speaking of which, the "Q1" on this row is quite a small hit box, and is the only way to access the Question page for that question.
> 
> Screenshot 2022-01-13 at 09 48 58
> 
> What do you think about making the _entire Question wording_ a link to the Question page?
> 
I wouldn't mind doing that, the reason why I change it and add it the link to the code is because, with the underline styling of the links the questions were looking really busy.

Other option is making the links in code for the question pop up a little bit more, and give them more padding as well.
> The `overflow-y: scroll` on `.question-content` is causing unnecessary scrollbars:
> 
> Screenshot 2022-01-13 at 09 38 23
> 
> While chaging the `overflow-y: scroll` to `overflow-y: auto` would fix this, it’d still leave people scrolling long question descriptions through a 200px tall letterbox.

> 
> My instinct is that we should truncate the descriptions to a manageable length (the first paragraph?) and have either a JavaScript-powered "More…" button to reveal the initially truncated section, or just a "More…" link that leads to the individual Question page for that question, where the full description would be shown.
> 
Let's go for the Javascript "More..."

> I’m not sure what these "X out of Y" values are meant to be telling me:
> 
> Screenshot 2022-01-13 at 09 44 16
> 
> I _assume_ it’s "X out of Y councils got full marks on this question", but we don’t explain that anywhere. Maybe we should? …
> 
> Screenshot 2022-01-13 at 09 47 04

zarino (approved):
Now this has a hero section at the top, I think we’re good to merge. I expect there will be some rough edges that need fixing, once @struan and I wire it up with real data next week.

PR #238, opened 2022-01-10
[WIP] sectoral emissions chart

### LINKED ISSUE #236: Display emissions data by sector, to give people an idea of the *sources* of emissions in an area
CAPE currently displays headline emissions statistics for each council, taken from [the 2005–2019 BEIS dataset](https://www.gov.uk/government/statistics/uk-local-authority-and-regional-carbon-dioxide-emissions-national-statistics-2005-to-2019):

Screenshot 2022-01-10 at 14 36 12

This data is fine, but doesn’t help visitors understand much about the profile of emissions in the area. It feels like there are three useful things we could help people understand:

1. How does this council’s emissions compare to other councils?
1. Where do these emissions come from?
1. How has this changed over time?

The BEIS dataset includes a sectoral breakdown for each council, each year:

* Grand total
   * Industry total
      * Electricity
      * Gas
      * Other fuels
      * Large industrial installations
      * Agriculture
   * Commercial total
      * Electricity
      * Gas
      * Other fuels
   * Public sector total
      * Electricity
      * Gas
      * Other fuels
   * Domestic total
      * Electricity
      * Gas
      * Other fuels
   * Transport total
      * Road transport (A roads)
      * Road transport (Minor roads)
      * Other

Exposing _some_ of this lower-level data would help visitors understand what _sort_ of emissions a council is dealing with in their area. From there, there’s a chance they might be able to understand _why_ certain decisions have been made in the council’s action plan, or assess _how much_ the council is doing to combat the most prevalent emissions sources in the area.

My instinct is the electricity/gas/other breakdowns are a level lower than most people would need. I wonder whether there’s a way we could exclude them in the default view, but let people drill down to see them, if they wanted?

Finally, I wondered whether replacing the sparkline (which looks the same for every council anyway) with a stacked line chart / area chart, might be a way to expose a lot of historical data in a single visualisation:

image

zarino:
Fixes #236.


zarino:
This has been superceded by amazing work by @ajparsons in https://github.com/mysociety/caps/pull/310.

PR #239, opened 2022-01-11
Updates to scoring to handle small tweaks to scoring import

struan:
You can test it by getting the CSV files from https://drive.google.com/drive/u/0/folders/1SzuHMfHu-yFhLx_gVigeDpa-jaysqJoH

Also includes the testing from previous PR


crowbot (changes requested):
Logic of the import looks good to me - a couple of oddities in looking at the actual data. On a council's own page, I see a couple of sections duplicated at the end of the list with zero scores
Screenshot 2022-01-11 at 19 06 57

Second thing is that (again on a council's page) the score and total for a couple of the sections don't seem to match the actual points available and scored - the total points listed as available in section Commitment and Integration is 7, but summing of the points available in each question of that section comes to 8. Similarly with Governance Development and Funding - the total says 21 points are available, but it looks like it's actually 22. For a sample council I looked at, the actual number of points scored looked one higher than the summary figure too for each of these sections - so e.g. listed as 16/21 where the totals would suggest it should be 17/22. Not sure if that's more for @ajparsons or you. I didn't clear out any old data before running `import_plan_scores` so let me know if I should have done that. 
The import process produced errors about failing to match the GLA - have we decided what we're going to do about that?

struan:
I think this might be partly an old data problem as I had this issue and then cleared the answers table and re-ran the import and it was fine. but also a display problem in that I was doing on left join on the questions table so pulling out questions even if there wasn't an answer so I will also fix that.

struan:
ok, that's done. The duplicate sections was another bug partly caused by old data but it should all be fixed now, although you might need to remove the bad data.

crowbot (approved):
👍

struan:
FWIW, this has been merged now

Issue #190, opened 2021-12-02
[CCL] Council page: Question table


zarino:
- [x] Row per section, subsection, and question. Columns for question number, question wording, this council’s score for the row, and the number of councils who got top score for the section/subsection (but not questions) – “x out of y” (with a link to show them all on the homepage?) (@lucascumsille / @struan)
- [x] JavaScript truncation for section, subsection, and question wording, with a “click to read more” link. (@zarino)
- [x] Section wording and “x out of y”  (@lucascumsille )
- [x] Subsection wording and “x out of y”. (@lucascumsille)
- [x] Show a star or something next to this council’s “full marks” scores (@lucascumsille)
- [x] Show a star or something next to this council’s “full marks” scores. ( @struan)
- [x] Add start in the "Top performers" head column(styling) (@lucascumsille)
- [ ] Add council to compare to question table (Hide on mobile) (@lucascumsille / @struan)

zarino:
We’re going to revisit the ability to add a council to compare at some future date.

PR #241, opened 2022-01-12
190 council page question fixed

lucascumsille:
Fixes part of: #190 

Screenshot 2022-01-12 at 14 03 27

Screenshot 2022-01-12 at 14 10 26


zarino (approved):
Looks fine – aside from the issues I mentioned in #237 (some of which I now realise I probably should have mentioned in _this_ PR, given they related to the question table on the _council_ page, and not the question table on the _question_ page, which is what that PR was about – oh well!)

PR #245, opened 2022-01-13
Scoring top performers

struan:
Calculate and save details of top performers when importing scores


crowbot (changes requested):
Code looks logical, but all the sections are starred as top performers for all councils - hard coded? 

There's another issue I just noticed which I don't think is introduced in this PR, which is that the sections and scores/total possible scores are matched in different ways in the tables and in the council answers page i.e. in the overall table, council x gets 4/7 for community engagement and communications and in the individual page it gets 9/10 - presumable something wonky in the way headings and scores are aligned in one of the places
Screenshot 2022-01-13 at 16 48 53
and in the individual table it gets 

Screenshot 2022-01-13 at 16 49 59

struan:
Front page/council page score mismatch fixed - the column header were incorrect on the front page.

I'm not seeing all the sections marked as top performer though.

crowbot (approved):
👍

PR #247, opened 2022-01-13
use authority type on front page

struan:
Highlight which authority type we are filtering by, and also change advanced filter options appropriately.

### DIFF
--- a/scoring/templates/scoring/includes/advanced-filter.html
+++ b/scoring/templates/scoring/includes/advanced-filter.html
@@ -6,6 +6,7 @@ data-popup-trigger="one" type="button">Advanced filters</button>
         <button class="close-icon popup-modal__close" type="button"></button>
     </div>
     <form>
+        {% if authority_type == 'single' %}
         <div class="criteria-group-wrapper">
             <button class="h5 btn-open-div" type="button">
                 Nation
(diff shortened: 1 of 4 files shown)


scoring/templates/scoring/includes/advanced-filter.html
```diff
@@ -6,6 +6,7 @@ <h4 class="filter-title">Select which criteria you would like to filter:</h4>
         <button class="close-icon popup-modal__close" type="button"></button>
     </div>
     <form>
+        {% if authority_type != 'northern-ireland' and authority_type != 'combined' %}
```
zarino, line 9: I _think_ (check with @ajparsons!) that you can add `district` and `county` to this list – ie: we only display the England/Scotland/Wales choice for Single Tier councils, because all of the districts and counties are in England.

zarino (approved):
Looks good, although I have left a question about whether we need to display the "Nation" options for District/County councils.

Also, I wonder whether we can now drop the temporary "Single Tier"/"District"/etc links from the bottom of the homepage template?

scoring/templates/scoring/includes/advanced-filter.html
ajparsons: I don't quite understand the issue, but yes, there are no district/county councils outside of England.

PR #252, opened 2022-01-14
Scoring import release based

ajparsons:
The original https://github.com/mysociety/caps/pull/239 is merged, but opening up a new one (I think the merge path is complicated at this point, so can just lift out the bits that are useful from this). 

This PR has two elements, the first switches the data upload from a manual upload to being based on github's release system. 

There is a [new repo](https://github.com/mysociety/climate_scorecard_data ) that contains just what will be the public data. This is generated directly from the processing repo, so is cutting down the number of files that have to move around. 

CAPE will pull down a specific release from this. This now includes the questions.csv so that doesn't need to be downloaded separately. Because this repo is private before we deploy live, this requires a PERSONAL_ACCESS_KEY environmental variable with a github personal access key with REPO permissions. When getting ready for live, the repo can be made public, and the private flag removed from the `settings.py`.

The second element is more of an optimisation, because there are so many individual answers being imported, there's a big speed gain in using `bulk_create` to import them.


scoring/management/commands/import_plan_scores.py
```diff
@@ -16,14 +23,63 @@
...
+    """
+
+    file = f"https://github.com/{org}/{repo}/archive/refs/tags/{tag}.zip"
+
+    headers = None
+    if private:
+        token = os.environ.get("PERSONAL_ACCESS_TOKEN", None)
+        if token is None:
+            raise ValueError(
+                "Tried to access private repo, but no PERSONAL_ACCESS_TOKEN envkey.")
```
struan, line 46: I think this might need to be set in the environment bit of `docker-compose.yml` in order that it is set when the container is started.

struan (changes requested):
One tiny thing but otherwise I think this looks good.

scoring/management/commands/import_plan_scores.py
ajparsons: Added below, I think this means it will now work that adding a token from https://github.com/settings/tokens to .env will then be pulled through into the docker environment. 

The live version shouldn't need this because the files will be open, but a staging site might need someone to give it a temporary token?

ajparsons: (Did some googling on this and annoyingly repo level tokens aren't available (despite codespaces clearly creating and using them behind the scenes).

ajparsons:
What branch should this be trying to merge into at this point?

struan:
council-climate-plan-scoring is the appropriate branch

ajparsons:
Ok, if good, that should be ok to merge.

Issue #216, opened 2021-12-22
Rename "Newcastle Upon Tyne, North Tyneside and Northumberland Combined Authority" to "North of Tyne Combined Authority"


zarino:
https://data.climateemergency.uk/councils/newcastle-upon-tyne-north-tyneside-and-northumberland-combined-authority/

They call themselves the "North of Tyne Combined Authority" now:

https://www.northoftyne-ca.gov.uk/

ajparsons:
This is correct in the new source data if that helps https://github.com/mysociety/uk_local_authority_names_and_codes/blob/main/data/uk_local_authorities.json#L12495

zarino:
@struan this still isn’t fixed. (ISTR we thought the name change would get picked up next time we refreshed the site? Clearly not.)

Also, "Sheffield City Region Combined Authority" renamed itself to "South Yorkshire Mayoral Combined Authority" in September 2021 – although that hasn’t made it into @ajparsons’ [uk_local_authorities.json](https://github.com/mysociety/uk_local_authority_names_and_codes/blob/main/data/uk_local_authorities.json) yet.

ajparsons:
I had a go at this and there's a few tricky areas:

* Renaming the councils in [this sheet](https://docs.google.com/spreadsheets/d/1tEnjJRaWsdXtCkMwA25-ZZ8D75zAY6c2GOOeUchZsnU/edit#gid=0) gets them into the database on a fresh load.
 * However it loses all climate promises and emergency declarations because these are listed in a different sheet. I think changing the name there should also fix this, but haven't tried it.
 * Also North of Tyne loses its plan (it looks like it's been deleted from that sheet?)
 
I've changed the names back in the main sheet for now. 

A practical issue before that is the council import code assume names do not change. I've fixed this in a PR, but this needs to be merged before trying to change any names. Even then, because we use a slug based on the name, not the authority code, in the url, the urls to old councils will disappear (I think this is fine, but fyi).

zarino:
Yes, I assumed the old council page URLs would disappear. Still, feels like the benefit of having nice human-readable council name slugs in the URLs outweighs the ocassional annoyance of council page URLs breaking when councils change their names. I guess the ideal situation would be for us to use the `alt-names` list in your JSON file to handle redirects when an unrecognised council slug is requested. Something for another day!

PR #255, opened 2022-01-14
Unique council loading

ajparsons:
Small fix: Fixed a current error when importing from council websites.

Bigger fix:  https://github.com/mysociety/caps/issues/216 requires changing a council's name, but currently that would lead to duplicate entries. This changes the council import process to just check for uniqueness on code and country.


struan (approved):
👍 

Should probably add some tests for this at some point.

Issue #251, opened 2022-01-14
GLA needs to be added to CAPE


ajparsons:
GLA plan is marked but GLA is not currently in the main cape database.

It needs to be added (just in general), but also so scorecard import can correctly load the GLA score.

ajparsons:
Adding it to the plan sheet gets it into the database:

![image](https://user-images.githubusercontent.com/8157058/149557197-5943d1c3-6124-4e85-8f8a-ebfd7bd48a3a.png)

Problems are then: 

1. Set for moment to combined authority - but might not be the best description. 
2. Needs details added on climate emergency declaration.

ajparsons:
GLA listed on relevant spreadsheets.

zarino:
I think this has been done now – https://data.climateemergency.uk/councils/greater-london-authority/

(Re-open if I’ve missed something!)

PR #257, opened 2022-01-17
List correct powers of Greater London Authority

ajparsons:
For importing the GLA: https://github.com/mysociety/caps/issues/251

This PR adds processing to list the correct powers of the GLA.

image

### DIFF
--- a/caps/models.py
+++ b/caps/models.py
@@ -187,17 +187,17 @@ class Council(models.Model):
         ):
             powers.append("passenger-transport")
 
-        if (
-            self.country != self.NORTHERN_IRELAND
-            and self.authority_type != "COMB"
-            and self.authority_type != "NMD"
-        ):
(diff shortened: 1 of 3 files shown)


caps/models.py
```diff
@@ -189,15 +189,15 @@ def powers(self):
 
         if (
             self.country != self.NORTHERN_IRELAND
-            and self.authority_type != "COMB"
+            and self.authority_type not in ["COMB", "SRA"]
```
struan, line 192: I realise there's some benefit in splitting this up by conceptually similar authority types but equally we could just add "NMD" to the list and only have two conditions here, and below and I think it would be fine.

struan (approved):
one tiny comment that you should feel free to ignore.

caps/models.py
ajparsons: Yep.

PR #258, opened 2022-01-17
Fix question display issues

struan:
Fix a few things with the display on the answer page.

This doesn't deal with the total in sub-sections.


crowbot (approved):
Looks good to me - the test changes don't match the code changes, so presumably are mopping up the fact that the tests weren't tracking code changes in the branch this is being merged into.

struan:
Yes, the test changes are sadly related to earlier breakages.

PR #259, opened 2022-01-17
Import council filter data

### LINKED ISSUE #248: Import council filter data
Use this data: https://github.com/mysociety/climate_league_filter_datasets/blob/main/data/outputs/all_filters.csv

struan:
Fixes #248.

Add model and management command to import the data required for scoring filters


crowbot (approved):
👍  Looks good.

ajparsons:
I'm not sure about adding stuff like `political_control` (which isn't constant and we're not planning to update) to the caps council model.  It adds an extra step later on, but could this be a model in the scoring app with a foreign key to the caps `Council`? Could also then have a 'year' variable to mirror the questions for some light future-proofing.

caps/management/commands/import_filter_data.py
```diff
@@ -0,0 +1,69 @@
...
+def get_data():
+    token = os.environ.get("PERSONAL_ACCESS_TOKEN", None)
+    headers = {"Authorization": "token " + token}
+    r = requests.get(FILTER_DATA_URL, headers=headers)
+    with open(FILTER_CSV, "wb") as outfile:
+        outfile.write(r.content)
+
+
+def import_filters():
+    ruc_lookup = dict((desc, code) for code, desc in Council.RUC_TYPES)
```
ajparsons, line 32: Doesn't matter, but any reason not to do `{desc: code for code, desc in Council.RUC_TYPES}` here?

caps/management/commands/import_filter_data.py
```diff
@@ -0,0 +1,69 @@
...
+
+    updates = []
+    df = pd.read_csv(FILTER_CSV)
+    for index, row in df.iterrows():
+        council = Council.objects.get(authority_code=row["local-authority-code"])
+
+        quintile = PlanDocument.integer_from_text(row["council-quintile"])
```
ajparsons, line 39: Doesn't need fixing, but could also be done as `df["council-quintile"] = df["council-quintile"].astype(int)` to force everything to int before going into the loop.

caps/management/commands/import_filter_data.py
```diff
@@ -0,0 +1,69 @@
...
+        if quintile is None:
+            quintile = 0
+
+        try:
+            cluster = ruc_lookup[row["ruc_cluster"]]
+        except KeyError:
+            cluster = ""
+
+        council.population = PlanDocument.char_from_text(row["pop_bucket"])
```
ajparsons, line 48: Similarly `row["pop_bucket"] = row["pop_bucket"].fillna("")` would be clearer about what was wrong with the column in the first place.

caps/management/commands/import_filter_data.py
ajparsons: Also `.fillna(0)` would take care of the next few lines.

caps/management/commands/import_filter_data.py
struan: no reason at all. Is there a general preference for one over the other?

ajparsons: not really, I think technically there's an efficiency thing because you're not making a separate object first, but it doesn't matter at this size.

struan:
On the political control thing, it might make more sense to attach that to the planscore as, yes, it's a property of the scoring not the council.

I will do that.

ajparsons:
> On the political control thing, it might make more sense to attach that to the planscore as, yes, it's a property of the scoring not the council.
> 
> I will do that.

This also applies a bit more generally to the deprivation stuff, as those quintiles are especially for scorecard groups (which aren't used in cape in the same way, and are not 'universal' quintiles).  Also population is missing for some council types because the filter didn't need them, and the buckets sizes are designed for the filter groups, so again might not be useful in this form ofr cape.

struan:
Ah, so, perhaps if I just add all these as properties of the plan score rather than the council that will make more sense?

ajparsons:
Yeah, I think that's a bit cleaner.

caps/management/commands/import_filter_data.py
```diff
@@ -0,0 +1,68 @@
...
+    ruc_lookup = dict((desc, code) for code, desc in PlanScore.RUC_TYPES)
+
+    updates = []
+    df = pd.read_csv(FILTER_CSV)
+    df["council-quintile"] = df["council-quintile"].fillna(0).astype(int)
+    df["pop_bucket"] = df["pop_bucket"].fillna("")
+    df["political_control"] = df["political_control"].fillna("")
+
+    for index, row in df.iterrows():
+        plan_score = PlanScore.objects.get(year=2021, council__authority_code=row["local-authority-code"])
```
ajparsons, line 42: In other cases I think you referred to a year variable in settings? But also, not a blocker.

ajparsons (commented):
Yep, that model approach looks better. I think at some point we should get all various external datasets being pulled in through a similar system, but it works and that's a later problem.

ajparsons:
^ Opps, did not mean to do that merge fix in this PR. Ignore if bad.

PR #266, opened 2022-01-19
“Jump to council” on homepage table

### LINKED ISSUE #181: [CCL] Homepage: Jump to council, by name
- [x] Search box at top of homepage(s)
- [x] JavaScript autocomplete – Display matching council names, once 3 or more characters are entered (like CAPE hompage) (@zarino)
- [ ] Picking a council switches to the correct homepage, and jumps you down to the council’s matching row in the table, highlighting the table row (@zarino)

zarino:
Fixes #181.

Not sure what the browser support is for all this JavaScript, but I figure it’s better than nothing, and we can revisit/polyfill support for older browsers if we get time, later on.

@lucascumsille could you check out this branch, and confirm you’re happy with how this looks and works?

### DIFF
--- a/caps/management/commands/import_filter_data.py
+++ b/caps/management/commands/import_filter_data.py
@@ -41,9 +41,17 @@ def import_filters():
     df["political_control"] = df["political_control"].fillna("")
 
     for index, row in df.iterrows():
-        plan_score = PlanScore.objects.get(
-            year=YEAR, council__authority_code=row["local-authority-code"]
-        )
+        try:
+            plan_score = PlanScore.objects.get(
+                year=YEAR, council__authority_code=row["local-authority-code"]
+            )
+        except PlanScore.DoesNotExist:
+            print(
+                "could not find plan score for council {}".format(
+                    row["local-authority-code"]
+                )
+            )
+            continue
 
         try:
             cluster = ruc_lookup[row["ruc_cluster"]]
--- a/proj/settings.py
+++ b/proj/settings.py
@@ -32,7 +32,7 @@ else:
     IS_LIVE = True
     STATICFILES_STORAGE = "pipeline.storage.PipelineStorage"
 
-ALLOWED_HOSTS = ["127.0.0.1", "testserver", "0.0.0.0", "localhost"]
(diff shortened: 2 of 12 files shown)


zarino:
Oh, damn, just realised this doesn’t yet handle the situation where you’re searching for a council that _isn’t_ on the current "page" – ie: you’re searching for a county council, when the table currently only contains single tier councils.

Hmmm… will have a think about how we can address that.

zarino:
@lucascumsille This is ready for you to try out now! Let me know what you think.

lucascumsille:
> @lucascumsille This is ready for you to try out now! Let me know what you think.

There is a problem with the background-color whenever you scroll more than once within the table.

https://user-images.githubusercontent.com/13790153/150336474-f8b24992-f1a7-422c-96e4-1fc427144916.mov

zarino:
@lucascumsille My intention was that, when you jump to a row, the background of the row turns strong yellow, then fades over the next few seconds, back to normal.

Do you think the background should stay yellow forever? (Or until you search for another council?)

lucascumsille:
> @lucascumsille My intention was that, when you jump to a row, the background of the row turns strong yellow, then fades over the next few seconds, back to normal.
> 
> Do you think the background should stay yellow forever? (Or until you search for another council?)
Good question, now that I think about it the table is not sorted out from highest to lowest anymore. Ideally the table should work in a way where I can sort out the table and create a "ranking" then I would look for the council that I'm searching for and get highlighted. Then I would scroll to see which councils are next to me, I probably would expect the highlight not to disappear.

Now that the table is not being sorted from highest to lowest, I'm not sure if keep it highlighted when scrolling is that important.

Finally I think that if I search for a new council using the search bar, yes I would expect only to highlight the new council and the old one should loose the highlight.

zarino:
@lucascumsille I’ve pushed a fixup commit that removes the fading behaviour, and leaves the row background yellow, until you either reload the page, or search for a new council. Seem ok to you?

lucascumsille (approved):
Looking great @zarino I think this makes more sense.

PR #267, opened 2022-01-19
display council data on council page

struan:
removed the budget as we don't have that data at the moment and then
moved political control down as it was smooshing up against the other
column where it was


zarino (approved):
@struan this looks good to me! I’ve added a fixup commit that hides sections we don’t have data for, eg on a Northern Irish council:

Screenshot 2022-01-20 at 17 18 49

If you’re happy with that fixup, you can go ahead and merge this.

PR #274, opened 2022-01-21
Added Contact page

lucascumsille:
Screenshot 2022-01-21 at 08 37 37


zarino (approved):
Again, looks like the wording might still be changing, but let’s get this merged in the meantime. :+1:

PR #285, opened 2022-01-24
Advanced filters

struan:


zarino:
@struan I’ve added some UI fixes. Figured I’d add a submit button in, for now, so that it at least works. We can attempt ajax later if we get time.

We need to make the radio buttons `checked` when they’re currently active (including the default "any"/"all" buttons when the page loads).

zarino:
Oh, derp, also, we need to display the current filter(s?) in this title, rather than it being hard-coded to "URBAN"

Screenshot 2022-01-25 at 10 24 23

scoring/templates/scoring/includes/advanced-filter.html
```diff
@@ -126,12 +126,12 @@
                 </legend>
                 <div class="option-group open-div">
                     <div class="radio-wrapper">
-                        <input type="radio" id="poulation_all" name="population" value="">
+                        <input type="radio" id="poulation_all" name="population" value=""{{ option.0 }}"{% if filter_params.population == "" or filter_params.population is None %} checked{% endif %}>
```
zarino, line 129: Typo: `value=""…"`

zarino: Or maybe the `{{ option.0 }}` shouldn’t be there?

PR #300, opened 2022-01-26
Fixed bug on safari (left scrolling)

lucascumsille:


zarino (approved):
Works here in Safari 15.2 :+1:

PR #301, opened 2022-01-26
add stategic regional authority to combined scoring table

struan:
this is to cover the GLA


struan:
Merged in b5439a1d058f2b738c92fc5f4253929ab27aa9ad

PR #310, opened 2022-01-28
Multi emissions chart

ajparsons:
In progress pull request for adding a chart about different emissions types to CAPE (#236).

In addition to the chart, I've added some quick text bringing back in the labels from the clustering approach (these might need a light rewording). 

This PR adds helpers for working with altair/vega charts. In addition to getting the chart into the view, these helpers do the following:

- When `VEGALITE_SERVER_URL` and `VEGALITE_ENCRYPT_KEY` are configured (See wiki passwords page), a fallback image (using https://github.com/mysociety/vegalite-server) is rendered when javascript is off or if ES6 is not supported. 
- The chart can be paired with a dataframe that makes a table of the data available for screen readers. 
- Extra export options add the logo and export in higher quality in a few aspect ratio:

The current result:

image

This chart has a tooltip:

image

image

image


caps/views.py
```diff
@@ -97,6 +103,19 @@ def get_context_data(self, **kwargs):
         except DataPoint.DoesNotExist:
             context["no_emissions_data"] = True
 
+        context["current_emissions_breakdown"] = council.current_emissions_breakdown()
```
ajparsons, line 106: think this bit needs to be be wrapped in a check for `no_emission_data` set above to get the test workin.

caps/charts.py
```diff
@@ -0,0 +1,92 @@
+from typing import TYPE_CHECKING
+
+from .models import Council, DataPoint
+
+import altair as alt
+import pandas as pd
+from charting import ChartBundle
+
+
+def multi_emissions_chart(council: "Council"):
```
ajparsons, line 10: Think can import the future annotations thing and take the Council out of the string

caps/charts.py
```diff
@@ -0,0 +1,92 @@
...
+        "Industry Total",
+        "Commercial Total",
+        "Public Sector Total",
+        "Transport Total",
+        "Domestic Total",
+    ]
+
+    df = (
+        DataPoint.objects.filter(council=council, data_type__name_in_source__in=totals)
+        .to_dataframe("year", "data_type__name_in_source", "value")
```
ajparsons, line 23: Could add something to `to_dataframe` to use tuples to rename rather than always doing this in two steps.

caps/charts.py
```diff
@@ -0,0 +1,92 @@
...
+        .reset_index()
+        .melt(id_vars=["year"])
+        .rename(columns={"value": "row_percentage"})
+    )
+    df = df.merge(pdf)
+
+    # Tidy Emissions type label
+    df["emissions_type"] = df["emissions_type"].str.replace(" Total", "")
+
+    # altair doens't pick up years right unless in a fake data format
```
ajparsons, line 40: 'date'

charting/control.py
```diff
@@ -0,0 +1,79 @@
...
+    To create multiple-rendering options.
+
+    """
+
+    label: str
+    alt_title: str
+    df: pd.DataFrame
+    chart: alt.Chart
+    spec: str = field(init=False)
+    ident: str = field(init=False)
```
ajparsons, line 28: move late initiated ones to end

charting/control.py
```diff
@@ -0,0 +1,79 @@
...
+    logo_src: str = "/static/charting/img/mysociety-logo-white-background.jpg"
+    data_source: str = ""
+
+    def __post_init__(self):
+        self.spec = self.chart.to_json()
+        self.ident = self.get_ident()
+        self.image_url = self.get_image_url()
```
ajparsons, line 35: image_url not explicit above

charting/control.py
```diff
@@ -0,0 +1,79 @@
...
+class ChartBundle:
+    """
+    The chart bundle contains the dataframe and the altair chart.
+    To create multiple-rendering options.
+
+    """
+
+    label: str
+    alt_title: str
+    df: pd.DataFrame
```
ajparsons, line 25: What's currently being returned by the chart is a styler. Both have similar `to_html' methods, so should be allowed as an option here.

charting/templates/charting/includes/chart-js.html
```diff
@@ -0,0 +1,22 @@
+var {{chart.ident}}_spec = {{chart.spec|safe}};
```
ajparsons, line 1: I don't *love* this approach, but it works.

charting/templates/charting/includes/chart-js.html
```diff
@@ -0,0 +1,22 @@
...
+    {{chart.ident}}_spec["width"] = Math.floor(facet_width);
+{% endif %}
+results = await vegaEmbed("#{{chart.ident}}", {{chart.ident}}_spec, embed_opt);
+const {{chart.ident}}_view = results.view;
+{{chart.ident}}_print_action = document.createElement('a');
+{{chart.ident}}_print_action.href = "#";
+{{chart.ident}}_print_action.textContent = "Copy (for print)";
+{{chart.ident}}_print_action.onclick = function(event) {toClipboardPrint({{chart.ident}}, {{chart.ident}}_view, {{chart.ident}}_spec,); event.preventDefault();};
+{{chart.ident}}.querySelector('.vega-actions').appendChild({{chart.ident}}_print_action)
+{{chart.ident}}_web_action = document.createElement('a');
```
ajparsons, line 18: Maybe these buttons should be moved out of the drop down to below the image to make them more obvious?

charting/templates/charting/includes/chart-render.html
```diff
@@ -0,0 +1,20 @@
+{% load static %}
+
+<script src="https://cdn.jsdelivr.net/npm/vega@5.9.0"></script>
```
ajparsons, line 3: Could be replaced by local versions

proj/settings.py
```diff
@@ -115,7 +115,7 @@
     },
     "JAVASCRIPT": {
         "caps": {
-            "source_filenames": ("caps/js/main.js",),
+            "source_filenames": ("caps/js/main.js", "charting/js/altair-fallback.js"),
```
ajparsons, line 118: Unsure if this is necessary - and has the wrong filename anyway.

caps/models.py
```diff
@@ -579,6 +607,14 @@ def __str__(self):
         return self.name
 
 
+class QuerySetWithDataFrame(models.QuerySet):
```
ajparsons, line 610: Could apply this to all classes for consistency.

charting/static/charting/js/altair_fallback.js
```diff
@@ -0,0 +1,122 @@
...
+        text_width = ctx.measureText(data_source + "   ").width;
+        ctx.fillText(data_source, new_canvas.width - text_width, new_canvas.height - 50);
+        base64data = new_canvas.toDataURL(1)
+
+        fetch(base64data)
+            .then(res => {
+                return res.blob();
+            })
+            .then(blob => {
+                navigator.clipboard.write([new ClipboardItem({ 'image/png': blob })]);
```
zarino, line 114: FWIW `ClipboardItem` doesn’t work in Firefox or IE. Probably not big enough user numbers to care (only about 2.5% of CAPE visitors) but does seem a shame that we present menu items for copying images, which do nothing when you click them in those browsers.

zarino: Maybe `chart-js.html` should check whether `window.ClipboardItem` exists, before appending those items to the menu?

charting/templates/charting/includes/chart-render.html
zarino: Yes, should deffo do this.

zarino (commented):
I’ve added a few fixup commits that:

1. Remove the emissions profile name / description – I want to try to keep this section simple for now, and it seems like emissions profiles are more useful when presented as a way of grouping councils, and less so as a way of _describing_ a council in isolation.
2. Add custom styling for the emissions data section.
3. Fix an error with the variable interpolation in the Docker config files.

charting/static/charting/js/altair_fallback.js
ajparsons: Done.

caps/models.py
ajparsons: Did this.

ajparsons:
@struan : Ready for you.

Some instructions up the top on new env variables, but otherwise should be hopefully straightforward.

caps/models.py
```diff
@@ -367,7 +426,7 @@ def get_available_name(self, name, max_length):
 overwrite_storage = OverwriteStorage()
 
 
-class PlanDocument(models.Model):
+class PlanDocument(CapsModel):
```
struan, line 429: do we need to replace the base model on every class? it looks like we only need it on the DataPoint class.

I don't suppose it does any harm but I think I have a slight preference for inheriting from standard classes unless there's a reason not to.

caps/views.py
```diff
@@ -95,8 +100,17 @@ def get_context_data(self, **kwargs):
             context["latest_year_per_km2_emissions"] = latest_year_per_km2_emissions
             context["latest_year_total_emissions"] = latest_year_total_emissions
         except DataPoint.DoesNotExist:
+            no_emissions_data = True
             context["no_emissions_data"] = True
 
+        if not no_emissions_data:
```
struan, line 106: I think this would be a bit clearer if this was `emissions_data` and the assignment earlier were reversed (i.e. defaults to True and False in exception.

charting/templates/charting/includes/chart-display.html
```diff
@@ -1,16 +1,16 @@
 <div class="chart-and-title">
-    {% if chart.title and chart.html_chart_titles %}
+  {% if chart.title and chart.html_chart_titles %}
```
struan, line 2: given we never use it is there a reason to check for `html_chart_titles`?

charting/templates/charting/includes/chart-display.html
```diff
@@ -1,16 +1,16 @@
...
     <img _src="{{chart.image_url|safe}}" class="es5" alt="Chart: {{chart.alt_title}}" longdesc="#longdesc_{{chart.ident}}" style="display:none">
     <noscript>
-    <img src="{{chart.image_url|safe}}" class="" alt="Chart: {{chart.alt_title}}" longdesc="#longdesc_{{chart.ident}}">
+        <img src="{{chart.image_url|safe}}" class="" alt="Chart: {{chart.alt_title}}" longdesc="#longdesc_{{chart.ident}}">
     </noscript>
-    </div>
+</div>
 
-    <div id="longdesc_{{chart.itent}}" class="longdesc" style="display:none">
+<div id="longdesc_{{chart.itent}}" class="longdesc" style="display:none">
```
struan, line 12: should this be `chart.ident`?

caps/charts.py
```diff
@@ -0,0 +1,110 @@
...
+        .reset_index()
+        .melt(id_vars=["year"])
+        .rename(columns={"value": "row_percentage"})
+    )
+    df = df.merge(pdf)
+
+    # Tidy Emissions type label
+    df["emissions_type"] = df["emissions_type"].str.replace(" Total", "")
+
+    # altair doens't pick up years right unless in a fake date format
```
struan, line 38: typo in comment

charting/control.py
```diff
@@ -0,0 +1,81 @@
...
+        ident = ""
+        for h in hash:
+            if h in string.digits:
+                ident += chr(65 + 8 + int(h))
+            else:
+                ident += h
+
+        return ident[:6]
+
+    def get_image_url(self) -> str:
```
struan, line 51: as discussed in slack this currently doesn't seem to work so possibly we should remove it and skip the chart if the browser has no JS/ES6 support

struan (changes requested):
There's a few small things, plus the fallback chart issues but otherwise this is nice.

proj/settings.py
```diff
@@ -119,7 +119,7 @@
...
```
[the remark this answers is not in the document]
zarino: Plus it would be good to only load this extra JS on pages that actually need it (rather than concatenating it all into the same `main.js` that gets loaded on every page).

caps/templates/council_detail.html
```diff
@@ -209,33 +209,46 @@ <h2>Climate Plan Scorecard</h2>
...
+                        <div>
+                            <span class="h3">{{ latest_year_per_km2_emissions|floatformat }}</span>
+                            <abbr title="Kilotonnes of Carbon Dioxide Equivalent">ktCO2e</abbr>
+                        </div>
                         <p>per km<sup>2</sup></p>
                     </div>
                 </div>
-
-                <p class="text-muted mb-0"><small>Latest CO<sub>2</sub> estimates from <a href="https://www.gov.uk/government/statistics/uk-local-authority-and-regional-carbon-dioxide-emissions-national-statistics-2005-to-2019">the Department of Business, Energy &amp; Industrial Strategy</a>. County and combined authority data combined from constituent authorities.</small></p>
-
```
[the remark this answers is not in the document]
zarino: Yeah, my fault, it got dropped at one point in my back-and-forthing over the layout, and never worked its way back in. The source for CO2 figures _is_ still mentioned on `/about/data`, but yeah, no reason not to work this back into the page now, below the chart, I assume.

charting/control.py
zarino: The alternative @ajparsons and I mentioned in passing earlier this week was to display a `<table>` of the historical sector data by default, and then replace the table with the chart when JS is available. Would that be much work to implement, @ajparsons?

caps/templates/council_detail.html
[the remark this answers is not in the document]
ajparsons: Will put it back.

caps/models.py
ajparsons: Sure thing.

charting/templates/charting/includes/chart-display.html
ajparsons: ...yes

charting/templates/charting/includes/chart-display.html
ajparsons: removed.

caps/templates/council_detail.html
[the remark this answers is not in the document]
ajparsons: image

struan (approved):
yay!

caps/views.py
ajparsons: This happened

PR #321, opened 2022-02-03
display and filter council answers on question page

### LINKED ISSUE #315: [CCL] Question page: Summary statistics about a single question or subpoint
eg: percentage of councils who scored full marks / zero marks on the question.

Screenshot 2022-02-03 at 14 15 53

The aim is that visitors should be able to get a general feel for how all councils performed on a single question, or subpoint, without having to visit many councils’ pages individually.

struan:
This adds the summary data to the single question page for both the question and the sub questions.

For this to work you need to run the migrations and then re-import the plan scores as it adds a parent property to the sub questions to make the totalling up easier:

```
./manage.py migrate
./manage.py import_plan_scores
```

(@zarino notes he did `docker-compose build app` and then ran `./manage.py import_plan_scores` from within a `docker-compose exec app bash` shell.)

Fixes #315


zarino (changes requested):
Mostly looks good to me. I‘ve pushed two commits that hide some of the non-functional stuff, and fix a nesting error in one of the for loops. @struan you’ll probably want to squash those into whichever of your original commits feels most relevant!

Before we merge this, I wonder whether @lucascumsille could take a look at this section of the question page, and sort out the huggledy-piggledy layout?

Screenshot 2022-02-08 at 18 05 10

lucascumsille:
> Mostly looks good to me. I‘ve pushed two commits that hide some of the non-functional stuff, and fix a nesting error in one of the for loops. @struan you’ll probably want to squash those into whichever of your original commits feels most relevant!
> 
> Before we merge this, I wonder whether @lucascumsille could take a look at this section of the question page, and sort out the huggledy-piggledy layout?
> 
> Screenshot 2022-02-08 at 18 05 10

@zarino I'll be working on it this morning. =)

lucascumsille:
@zarino and @struan I just pushed the fixes that Zarino mentioned. Let me know if you like the changes.

Screenshot 2022-02-09 at 08 16 25

Btw I noticed one bug, when opening the accordion and only for the header question. There is an error with Z% of council scored X out of Y. See the screenshot
Screenshot 2022-02-09 at 08 11 25

The subpoint questions are working alright.

struan:
@zarino the filtering etc works so there's no need to remove it.

struan:
I vote for hiding it for now as then we can get this page live.

lucascumsille:
@zarino The layout will remain the same. I would use the same highlight effect you use on the scorecards(yellow highlight), it would be cool if the accordion for the header question opens when the user search.

In summary:
1) The user search for a council
2) Opens accordion for header question
3) Yellow Highlight for the header question + subpoint questions. Though the only one that will be visible straight away would be for the header question, because the rest of the accordions items will be closed.

If this is is too difficult then:
When the user search for a new council, we don't display the accordion, just a table with with two columns (Question code and question text + on the second column points obtained/Total points). We would display the header question and the subpoint questions.

zarino (approved):
With the search box hidden, this all looks good to go :+1:

zarino:
Thanks @crowbot. I’ve added a commit that hides the stats if 0 councils match the given filters:

Screenshot 2022-02-11 at 15 15 37

PR #322, opened 2022-02-07
update english council url scraping

struan:
The LGA have changed the page to use a dl rather than a table, and also
removed all the twitter handles. This updates the parser to deal with
that, and also to only update the twitter handle if it's not blank.


crowbot (approved):
Looks good to me!

PR #324, opened 2022-02-07
Fix combined emissions

### LINKED ISSUE #323: CAPE emissions calculations for Combined Authorities are wrong
We’re calculating Combined Authority emissions totals incorrectly.

Eg: The "13.3 tCO2e per capita" we display [on North of Tyne Combined Authority’s CAPE page](https://data.climateemergency.uk/councils/north-of-tyne-combined-authority/) is wrong.

Looking at the BEIS subset data sheet, our 13.3 tonne figure appears to be the sum of the per-capita figures for the three constituent local authorities…

    ( 3.98 + 3.97 + 5.37 ) = 13.32 tonnes

…when it should really be the total emissions for the three authorities divided by the total population of the three authorities…

    ( 1205.3 + 825.3 + 1730.4 ) / ( 302820 + 207913 + 322434 ) = 0.0045 kilotonnes = 4.5 tonnes

I think the same bug is also present in the "20.5 ktCO2e per km<sup>2</sup>" figure – looks to me like a sum of the constituent local authorities’ figures, rather than total emissions divided by total area.

Also, while we’re at it – are we sure these figures are really about CO<sub>2</sub> _equivalent_? The BEIS spreadsheet only mentions CO<sub>2</sub>.

ajparsons:
Fixed #323. 

* Fixed bug in calculating per person and per area emissions for higher geographies.
* Improved wording on display of emissions data.

image

Needs to run `python manage.py import_emissions_data --all` after merge.

### DIFF
--- a/caps/management/commands/add_council_websites.py
+++ b/caps/management/commands/add_council_websites.py
@@ -41,24 +41,18 @@ def get_england_and_wales():
         print("problem getting England and Wales URL data: {}".format(e))
         return []
 
-    table = dom.find("div", class_="table-responsive")
-    body = table.find("tbody")
-    rows = body.find_all("tr")
+    council_list = dom.find("dl", class_="ckeditor-accordion")
+    rows = council_list.find_all("p")
     for row in rows:
-        link = row.find("th").find("a")
+        link = row.find("a")
         if link is None:
             continue
         council = link.text
         website_url = link["href"]
-        twitter_link = row.find("td").find(href=re.compile(r"twitter.com"))
-        twitter_url = twitter_link["href"]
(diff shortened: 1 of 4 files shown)


zarino (approved):
Wording changes look good to me.

I assume you’ve run this locally and confirmed that removing the `(t)` and `(kt)` from the column names has fixed the bug that was causing the incorrect combined authority totals in #323?

ajparsons:
@zarino : Yep, that's the screenshot from the new data.

ajparsons:
@struan If you're running any updates tomorrow, can you also merge and rerun the import command?

PR #329, opened 2022-02-10
Add header question totals, councils with max score counts to council page

### LINKED ISSUE #316: [CCL] Council page: “X out of Y councils got full marks” column for each question in the Questions table
Only required for questions, not subpoints. In the mockups, this looked like:

Screenshot 2022-02-03 at 14 21 24

(Since the mockups were produced, the intended display changed a bit, to make it clear that this final column is about _councils who got full marks_ – all the markup should be in the git history.)

Remember, we should only be showing data for other councils of the same `authority_type`.

### LINKED ISSUE #318: [CCL] Council page: Display totals for questions (ie: “headings”) rather than just the scores for their constituent subpoints

Screenshot 2022-02-03 at 14 44 27

(Note the "5/5" score for Doncaster, in row S1.1)

struan:
This adds the following to the council page:

* totals for header questions (e.g s1_gov_q1)
* counts of councils with maximum scores to sections and questions
* does a bit of refactoring of querysets out to models for tidyness

You need to run migrations and then re-run `import_plan_scores` in order for the header question scores to work.

Fixes #318 
Fixes #316

### DIFF
--- a/scoring/templates/scoring/council.html
+++ b/scoring/templates/scoring/council.html
@@ -27,7 +27,7 @@
             <div class="about-section-child pt-3">
                 <h3 class="mb-3">About this council</h3>
                 {% if plan_score.top_performer %}
-                <p class="top-performer.active mb-0">This council is a top performer within {{ authority_type.name }} councils.</p>
+                <p class="top-performer.active mb-0">This council is a top performer within {% include 'includes/authority_type.html' with group=authority_type %} councils.</p>
                 {% endif %}
                 <a href="{% url 'scoring' authority_type.slug %}">See council in context</a>
                 <div class="group-label-wrapper mt-3">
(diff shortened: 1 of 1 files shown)


zarino (changes requested):
Woop!

Generally all good. I’ve added a commit that tidies up the styles a bit, and also adds some ALL_CAPS pointers to missing bits in [`scoring/council.html`](https://github.com/mysociety/caps/pull/329/files#diff-1e2653a77a63462a63ba2c5b3aeff318bc710827f6c6672359b91d2b93596945). Could you sort those out? Once they’re in, I think this is good to go.

zarino (approved):
Looks good! I’ve added a fixup commit that lowercases `authority_type.name` so that the cell content still reads like a single sentence.

Think we’re good to squash and merge away!

zarino:
Damn, just spotted this will suffer from the same bug I spotted in https://github.com/mysociety/caps/pull/338 – namely that `{{ authority_type.name|lower }}` looks weird when `authority_type.name` == "Northern Ireland", which becomes "northern ireland" 🤔

PR #334, opened 2022-02-15
display the django debug toolbar when in debug mode

struan:


zarino (commented):
Would be great to have a way to turn _off_ the DjDT _even in `DEBUG` mode_, for those of us doing front-end design work / taking screenshots / etc, who don’t want the button getting in the way. 🙏

struan:
Good point, I've pushed a fixup that lets you add `HIDE_DEBUG_TOOLBAR=1` to your .env file which will hide it. You will need to restart the docker containter if you change it for it to take effect.

zarino (approved):
👍 Works!

PR #336, opened 2022-02-16
Add ability to compare councils to the council page

### LINKED ISSUE #317: [CCL] Council page: Add councils to compare, as columns on the Questions table

Screenshot 2022-02-03 at 14 27 02

And clicking the "Choose a council to compare" button revealed a pop-up like this, where you could see the councils already added to compare, search for a new council:

Prototype add councils to compare

We discussed displaying "suggestions" of councils to pick, eg: nearby councils – but I’m not sure we ever confirmed exactly _what_ types of councils we’d want to suggest. So let’s ignore that for now, and we can create another ticket for if it’s required.

In the meantime, since this is quite a complex bit of work, here’s the full spec:

* Clicking the button will reveal the popup
* Clicking the close button on the popup will close the popup
* Searching for a council via the search box in the popup, and selecting them, will add them (as a new column in the table, and also shown as "selected" in the popup)
   * Limit searches to just councils of the same `authority_type`
   * Limit number of councils that can be added to 3 – display a message when you reach the maximum
* Clicking the "x" button next to a council’s name in the popup, to remove them (which removes their column from the table, and removes them as "selected" in the popup)
* None of this needs to be visible or work in the "mobile view"

struan:
Enables comparisons at a section and header question level.

Fixes #317

### DIFF
--- a/scoring/models.py
+++ b/scoring/models.py
@@ -118,6 +118,22 @@ class PlanScore(models.Model):
         return questions
 
     @classmethod
+    def questions_answered_for_councils(cls, plan_ids=None, plan_year=None):
+        # do this in raw SQL as otherwise we need an extra query
+        questions = PlanQuestion.objects.raw(
+            "select q.id, q.code, q.text, q.question_type, q.max_score, s.code as section_code, a.answer, a.score, a.max_score as header_max \
+            from scoring_planquestion q join scoring_plansection s on q.section_id = s.id \
+            left join scoring_planquestionscore a on q.id = a.plan_question_id \
+            join scoring_planscore ps on a.plan_score_id = ps.id \
+            join caps_council c on ps.council_id = c.id \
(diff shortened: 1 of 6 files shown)


lucascumsille:
> @lucascumsille if you get a chance over the next few days, could you check out this branch, visit a council’s page, add a few other councils to compare, and then fix the missing borders in both of the tables on the page?

@zarino That's actually on purpose, in the original design only the current council have borders, it was a way to make it different from the others ones. As the the table columns have enough padding between each other, I didn't want to add borders to them. It felt. like noise noise. What you reckon, would you still want me to add it?

zarino:
> @zarino That's actually on purpose

Oh! Ok! Looked like a bug to me, but maybe normal people wouldn’t even notice. I’m not that worried about it, if you’re not.

PR #338, opened 2022-02-17
correctly determine group type for northern ireland councils

### LINKED ISSUE #337: Northern Irish council scorecard results (on CAPE) are compared to “Average single tier council” rather than “average Northern Irish council”
Example from [Derry City and Strabane District](https://data.climateemergency.uk/councils/derry-city-and-strabane-district-council/):

Screenshot 2022-02-17 at 11 18 25

Shouldn’t be comparing their score to the “Average single tier council” – should be comparing to the average Northern Ireland council.

struan:
Fixes #337

### DIFF
--- a/caps/models.py
+++ b/caps/models.py
@@ -268,14 +268,14 @@ class Council(models.Model):
         return df
 
     def get_scoring_group(self):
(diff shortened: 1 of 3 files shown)


zarino (approved):
Fab! 👍

PR #339, opened 2022-02-17
always use homepage for links to single authority scoring list

### LINKED ISSUE #330: [CCL] Homepage: Consolidate `/` and `/scoring/single` into one URL
Currently the same content (the "single tier councils" table) is available through two competing URLs:

* https://councilclimatescorecards.uk/
* https://councilclimatescorecards.uk/scoring/single/

We should either:

1. Make it so that `{% url "scoring" "single" %}` returns `/` rather than `/scoring/single` (thereby removing the `/scoring/single` URL entirely); or…
2. Set a `<link rel="canonical">` on one of the URLs, pointing to the other one, so that search engines know they’re duplicates of each other

struan:
Helps with search engine things

Fixes #330

### DIFF
--- a/scoring/templates/scoring/council.html
+++ b/scoring/templates/scoring/council.html
@@ -29,7 +29,7 @@
                 {% if plan_score.top_performer %}
                 <p class="top-performer.active mb-0">This council is a top performer within {% include 'includes/authority_type.html' with group=authority_type %} councils.</p>
                 {% endif %}
-                <a href="{% url 'scoring' authority_type.slug %}">See council in context</a>
+                <a href="{% include 'scoring/includes/scoring_url.html' with slug=council.get_scoring_group.slug %}">See council in context</a>
                 <div class="group-label-wrapper mt-3">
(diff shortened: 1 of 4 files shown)


zarino (approved):
This looks good to me. I made a few changes, @struan it’d be good if you could check you’re happy with them before merging:

1. I rebased the branch onto the latest `master`, so that we can also fix the homepage links added by #329
2. …although this required me to also remove the newline from the end of `scoring_url.html` because otherwise we couldn’t immediately follow the URL with the `?sort_by…` query string. (It’s really crappy that Django has such poor options for whitespace stripping around tags/includes.)
2. And I made it so that `scoring_url.html` takes a `slug` parameter (as a string), rather than a whole `authority_type` object (from which it would just pull `.slug` anyway) – as this allowed me to pass it a string in `main-filter.html` where we have the hard-coded homepage links, meaning we now no longer call `{% url 'scoring' … %}` on its own anywhere, we _always_ use the new partial. Yay!

PR #340, opened 2022-02-22
Import scottish council climate reporting data

struan:
This is an attempt to pull out some useful data from the mandatory scottish council climate reporting data.

To generate the data first create a `scottish` directory under data and then run:

`./manage.py fetch_scottish_reporting_data`

followed by

`./manage.py process_scottish_reporting_data`

this will download a load of excel files in the the `data\scottish` directory and then generate a number of CSV files in there with the parse out data.

if you run `process_scottish_reporting_data` you will see that you can limit the types processes which should speed things up a little if you don't want to process it all

### DIFF
--- a/caps/management/commands/process_scottish_reporting_data.py
+++ b/caps/management/commands/process_scottish_reporting_data.py
@@ -22,6 +22,7 @@ class Command(BaseCommand):
         "council_data": "basic council data",
         "emissions": "emissions data",
         "sources": "emissions sources data",
+        "targets": "reduction targets data",
     }
 
     def add_arguments(self, parser):
@@ -115,7 +116,15 @@ class Command(BaseCommand):
                 fetched = self.extract_emissions_sources_from_sheet()
                 parsed["sources"] = fetched
 
-    def make_data_point(self, point_type, point_value, sub_type="", units="", scope=""):
+            if (self.options["all"] or self.options["targets"]) and parsed.get(
+                "targets", 0
+            ) != 1:
+                fetched = self.extract_targets_from_sheet()
+                parsed["targets"] = fetched
+
+    def make_data_point(
+        self, point_type, point_value, sub_type="", units="", scope="", target_year=""
+    ):
         data_point = {
             "council": self.report_data["council"],
             "authority_code": self.report_data["authority_code"],
@@ -126,6 +135,7 @@ class Command(BaseCommand):
             "data_value": point_value,
             "units": units,
             "scope": scope,
+            "target_year": target_year,
         }
 
         return data_point
@@ -269,6 +279,46 @@ class Command(BaseCommand):
                 )
             )
 
+        return 0
+
+    def extract_targets_from_sheet(self):
+        column = self.get_description_column()
+        if column == -1:
+            return 0
+
+        try:
+            targets_df = self.get_row_range(column, r"Name of [Tt]arget")
+            if targets_df is not None:
+                type_header = "Name of Target"
+                if type_header not in targets_df.columns:
+                    type_header = "Name of target"
+
+                for index, row in targets_df.iterrows():
+                    point_type = row[type_header]
+                    point_data = row["Target"]
+                    point_units = row["Units"]
+                    year = row["Target completion year"]
+                    if point_data in [
+                        "Please select from drop down box",
+                        "Other (please specify in comments)",
+                    ]:
+                        continue
+                    data_point = self.make_data_point(
+                        "target",
+                        point_data,
+                        sub_type=point_type,
+                        units=point_units,
+                        target_year=year,
+                    )
+                    self.data.append(data_point)
+
+                return 1
+        except Exception as e:
+            print(
+                "problem parsing sheet {} for targets in {}: {}".format(
+                    self.sheet_name, self.report, e
+                )
+            )
         return 0
 
     def save_data(self):
--- a/caps/management/commands/process_scottish_reporting_data.py
+++ b/caps/management/commands/process_scottish_reporting_data.py
@@ -19,10 +19,26 @@ class Command(BaseCommand):
     report_data = None
 
     subsets = {
-        "council_data": "basic council data",
-        "emissions": "emissions data",
-        "sources": "emissions sources data",
-        "targets": "reduction targets data",
+        "council_data": {
+            "desc": "basic council data",
+            "method": "extract_council_data_from_sheet",
+        },
+        "emissions": {
+            "desc": "emissions data",
+            "method": "extract_emissions_data_from_sheet",
+            "header_text": "Baseline Year",
+            "start_text": "Reference year",
+        },
+        "sources": {
+            "desc": "emissions sources data",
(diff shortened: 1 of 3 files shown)


ajparsons:
I guess my meta question on this (and relates retrospectively to other data) is if this kind of manipulation belongs *outside* cape, and cape just pulls in processed versions?

struan:
That is a fair question. It would certainly make it a bit easier for other people to use. I guess you'd envisage something like the emissions data where there's another repo that does the processing and generation and CAPE fetches the data from that?

I think we can always move it outside at a later date. I plan to write anything to pull the data in as a third management command regardless which will make that easier.

ajparsons:
Yeah, especially where there's also some scraping, feels good to insulate it.

caps/management/commands/process_scottish_reporting_data.py
```diff
@@ -0,0 +1,428 @@
...
+                )
+            )
+            return
+
+        self.report = join(settings.SCOTTISH_DIR, self.report_data["file"])
+
+        xls = pd.ExcelFile(self.report)
+        parsed = {}
+        for sheet_name in xls.sheet_names:
+            self.sheet_name = sheet_name
```
ajparsons, line 124: Think it is working fine here, but this repeated assignment to a self makes me nervous in case something is missed and it comes around on next loop. If re-factoring at some point would probably move the sheet processing into its own class that can be loaded and destroyed?

caps/management/commands/process_scottish_reporting_data.py
```diff
@@ -0,0 +1,428 @@
...
+            "Are these savings figures estimated or actual? ",
+            "Are these savings figures estimated or actual?",
+            "Savings figures are estimated or actual ",
+            "Savings figures are estimated or actual",
+        ]
+
+        measurement = "Savings figures are estimated or actual"
+        for col in measurement_columns:
+            if col in targets_df.columns:
+                measurement = col
```
ajparsons, line 339: Think this bit could catch us out in future if another one of these variants appears.
Could add line 336 to the big list and reframe as:

valid_cols = [x for x in targets_df.columns if x in measurement_columns]
assert len(valid_cols) == 1
measurement = valid_cols[0]

caps/management/commands/process_scottish_reporting_data.py
```diff
@@ -0,0 +1,428 @@
...
+                    )
+
+            return 1
+
+    def save_data(self):
+        for name, data in self.data.items():
+            df = pd.DataFrame(data)
+
+            outfile = join(settings.SCOTTISH_DIR, "{}_data.csv".format(name))
+            df.to_csv(open(outfile, "w"), index=False, header=True)
```
ajparsons, line 428: [it works, so who cares] Any reason to give it the file object rather than just the string?

caps/management/commands/process_scottish_reporting_data.py
```diff
@@ -0,0 +1,428 @@
...
+
+    def process_data(self):
+        data = join(settings.SCOTTISH_DIR, "data_list.csv")
+        df = pd.read_csv(data)
+
+        for index, row in df.iterrows():
+            try:
+                self.report_data = row
+                self.process_report()
+                if self.options["sample"] and index == 20:
```
ajparsons, line 99: [comment, it's fine] this could also be
```
pd.read_csv(data, limit = 20 if self.options["sample"] else None)
```

or a less annoying way of setting the number

caps/management/commands/process_scottish_reporting_data.py
```diff
@@ -0,0 +1,428 @@
...
+        self.save_data()
+
+    def process_data(self):
+        data = join(settings.SCOTTISH_DIR, "data_list.csv")
+        df = pd.read_csv(data)
+
+        for index, row in df.iterrows():
+            try:
+                self.report_data = row
+                self.process_report()
```
ajparsons, line 98: [again, it works] Similar to the other self comment, I think it'd be better to pass row to `process_report` and have it assigned in there to self.report_data (so it's clear that process_report is effectively doing work on the row.)

ajparsons: (this also means you could replace the loop with df.apply(self.process_report), but no massive benefit to doing so in this case).

caps/management/commands/process_scottish_reporting_data.py
```diff
@@ -0,0 +1,428 @@
...
+                annual_savings=row["Estimated costs savings (£/annum)"],
+                measurement=row[measurement],
+                savings_start=row["First full year of CO2e savings"],
+                comments=row["Comments"],
+            )
+
+        return 1
+
+    def extract_generation_from_sheet(self, df):
+        if len(df.columns) == 3:
```
ajparsons, line 374: What leads to this being needed?

caps/management/commands/process_scottish_reporting_data.py
```diff
@@ -0,0 +1,428 @@
...
+                df.columns = columns
+
+                data = {}
+                for index, row in df.iterrows():
+                    if row["Technology"] == "Please select from drop down box":
+                        break
+
+                    consumed = 0
+                    exported = 0
+                    for v in row["Total consumed by the body (kWh)"].values:
```
ajparsons, line 406: Could you add a comment here explaining why there are multiple values and you prefer the last?

caps/management/commands/process_scottish_reporting_data.py
```diff
@@ -0,0 +1,428 @@
...
+        for col in measurement_columns:
+            if col in targets_df.columns:
+                measurement = col
+
+        for col in [
+            "Primary fuel/emission source saved",
+            measurement,
+            "First full year of CO2e savings",
+        ]:
+            targets_df[col] = targets_df[col].replace(
```
ajparsons, line 346: Don't do this, but fyi you could set this as an NA value on import and then use normal NA fill or processing.

```
read_csv(..., na_values = ["Please select from drop down box"])
```

caps/management/commands/process_scottish_reporting_data.py
```diff
@@ -0,0 +1,428 @@
...
+                    break
+            except Exception as e:
+                print(
+                    "problem processing report {} for council {}: {}".format(
+                        row["file"], row["council"], e
+                    )
+                )
+
+    def process_report(self):
+        # if data["start_year"] != 2020 and data["start_year"] != 2021:
```
ajparsons, line 109: Can come out now?

caps/management/commands/process_scottish_reporting_data.py
struan: largely because this is the pattern we're using elsewhere and might as well be consistent. I have no actual preference and the pandas docs don't seem to suggest one over the other.

caps/management/commands/process_scottish_reporting_data.py
struan: sigh, I think I did this before going away for a bit so only did half of it. I've done the rest now in a slightly better way with an explanation of what is going on.

PR #349, opened 2022-03-10
341 only count plans as plans

### LINKED ISSUE #341: Councils with only a pre-plan / meeting minutes etc, shouldn’t be listed as “having a plan”
From Isaac at CEUK:

> We list Councils that don't have a plan (only having a pre-plan/webpage) as having one.
>
> Can we mark these Councils as having "pre-plans" on the Council list page (https://data.climateemergency.uk/councils/), rather than "Yes :white_check_mark:" and remove them from the headline figure on the homepage (83% of local authorities have a plan).

struan:
Fixes #341

### DIFF
--- a/caps/templates/home.html
+++ b/caps/templates/home.html
@@ -22,7 +22,7 @@
                     {% include 'icons/chevron-right.html' with classes='ml-1' role='presentation' %}
                 </a>
                 <a data-show-feedback-modal href="{{ MEDIA_URL }}data/plans/plans.zip" class="d-flex align-items-center mt-3">
-                    Download ZIP archive of {{ total_plans }} plans ({{ plan_zip_size }})
+                    Download ZIP archive of {{ total_plans }} documents ({{ plan_zip_size }})
                     {% include 'icons/download.html' with classes='ml-1' role='presentation' %}
                 </a>
                 <div class="mt-3">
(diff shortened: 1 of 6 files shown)


zarino (approved):
This looks good from a front-end point of view. I fixed up the final commit, to prevent the line on the homepage overrunning.

Not sure I can comment on whether the Python code actually does what it should!

caps/templates/home.html
```diff
@@ -22,8 +22,8 @@ <h2 class="mb-0">{{ percent_councils_with_plan }}% of councils have a plan</h2>
                     {% include 'icons/chevron-right.html' with classes='ml-1' role='presentation' %}
                 </a>
                 <a data-show-feedback-modal href="{{ MEDIA_URL }}data/plans/plans.zip" class="d-flex align-items-center mt-3">
-                    Download ZIP archive of {{ total_plans }} plans ({{ plan_zip_size }})
-                    {% include 'icons/download.html' with classes='ml-1' role='presentation' %}
+                    Download ZIP archive of {{ total_plans }} documents ({{ plan_zip_size }})
+                    {% include 'icons/download.html' with classes='ml-1' %}
```
zarino, line 26: Any reason you got rid of the `role` here? We have `role="presentation"` set on our icons to tell assistive devices that they don’t need to read out the contents of the SVG (which in this case, is nothing, but it doesn’t _hurt_ to have it in there all the time.)

zarino (approved):
Latest changes look 💯 🎉  Love those tasty tests.

Issue #355, opened 2022-03-23
Accessibility fixes for CAPE


zarino:
Main issues are:

- Poor contrast when some of the primary colours are used against white (eg: blue text on white backgrounds, white text in orange buttons)
- ~`svg` elements should either have a label or, if they’re not meaningful on their own, `role="presentation"` to take them out of the accessibility tree~ – fixed by https://github.com/mysociety/caps/pull/356

zarino:
@lucascumsille fixed the most egregious colour contrast issues in https://github.com/mysociety/caps/pull/393. Think we can consider this ticket closed for now, and re-open if we spot any other showstoppers.

PR #356, opened 2022-03-23
Add role="presentation" to decorative SVG elements

zarino:
Fixes part of #355.

Been meaning to do this for a while!


struan (approved):
I wonder about all these very similar icon templates but I suspect refactoring it out to a base template is maybe not worthwhile given how rarely we change them.

caps/templates/includes/location-search-form.html
```diff
@@ -6,15 +6,15 @@
         </div>
         <div class="col-auto flex-shrink-0 d-flex">
             <button type="submit" class="btn btn-orange text-white btn-lg d-flex align-items-center">
-                {% include 'icons/search.html' %}
+                {% include 'icons/search.html' role='presentation' %}
```
struan, line 9: this is missing a with

caps/templates/includes/text-search-form.html
```diff
@@ -12,7 +12,7 @@
         </div>
         <div class="col-auto flex-shrink-0 d-flex">
             <button type="submit" class="btn btn-purple btn-lg d-flex align-items-center">
-                {% include 'icons/search.html' %}
+                {% include 'icons/search.html' role='presentation' %}
```
struan, line 15: also here

struan (changes requested):
oops, missed some problems

PR #358, opened 2022-03-24
Add section on telling us about data use

### LINKED ISSUE #346: Add option for users to provide email address when downloading ZIP archive from CAPE
Invite users to provide their email address when selecting **Download ZIP archive** from https://data.climateemergency.uk/ so that we can contact them to ask if they'd like to tell us how they've used the data, so that we can understand impact.

struan:
Fixes #346

### DIFF
--- a/caps/templates/about_data.html
+++ b/caps/templates/about_data.html
@@ -12,6 +12,10 @@
     </div>
     <div class="col-md-8 offset-lg-1">
 
+        <h2 class="mb-4" id="feedback">Feedback</h2>
+
+        <p>If you find this data useful we'd love to <a href="mailto:climate-councils@mysociety.org">hear from you</a> about how it was used. It's helpful for guiding future work, providing examples to others and for talking to existing and current funders.</p>
+
         <h2 class="mb-4" id="plans">Council plans</h2>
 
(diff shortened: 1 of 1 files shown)


caps/templates/about_data.html
```diff
@@ -12,6 +12,10 @@ <h1 class="mb-4 mb-lg-5">Our data</h1>
     </div>
     <div class="col-md-8 offset-lg-1">
 
+        <h2 class="mb-4" id="feedback">Feedback</h2>
+
+        <p>If you find this data useful we'd love to <a href="mailto:climate-councils@mysociety.org">hear from you</a> about how it was used. It's helpful for guiding future work, providing examples to others and for talking to existing and current funders.</p>
```
zarino, line 17: Curly quotes? :-D

zarino (approved):
How weird, could have sworn I’ve seen this code before. Maybe it was part of a different branch that never got merged.

My only thought is whether this should go on `/about` instead of `/about/data`. Can see an argument both ways. But there’s already a call for feedback on `/about`, and maybe it makes sense to have another one on `/about/data` for the folks who are looking up how our data works (most likely to be people actually _using_ the data).

PR #366, opened 2022-03-30
data download feedback request popup 

### LINKED ISSUE #346: Add option for users to provide email address when downloading ZIP archive from CAPE

struan:
Adds a modal popup to download links asking for an email address to write to about how data is being used.

Fixes #346

Screenshot 2022-03-31 at 16 42 04

### DIFF
--- a/caps/static/caps/js/main.js
+++ b/caps/static/caps/js/main.js
@@ -142,6 +142,29 @@ $('#interstitial-audience-survey').on('change', 'input', function(e){
     });
 });
 
+$('.data-download').on('click', function(e){
+    e.preventDefault();
+    $('#data-downloaded').val(e.currentTarget.href);
+    $('#data-feedback-modal').modal('show');
+});
+
+$('#data-feedback-modal').on('hide.bs.modal', function(e){
+    window.location = $('#data-downloaded').val();
+});
+
+$('form[data-ajax-feedback-submit]').on('submit', function(e){
+    e.preventDefault();
+    var $form = $(this);
+
+    $form.parents('.modal').modal('hide');
+
+    $.ajax({
+        method: $form.attr('method'),
+        url: $form.attr('action'),
+        data: $form.serialize()
+    });
+});
+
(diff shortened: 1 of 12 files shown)


zarino (changes requested):
@struan I‘ve added some styling for this modal, plus a bit of javascript that "enables" the download button when you’ve entered an email address (a suggestion from @MyfanwyNixon and @ajparsons that I think helps make it feel more like you’ve completed the requirements of the form).

Screenshot 2022-04-05 at 18 26 23

Two things I noticed:

1. This modal should probably also be triggered when you try to download the ZIP archive on the homepage.
2. The download should probably _not_ continue if you `esc` key out of the modal, or click the darkened background behind the modal to close it. (I get what you were going for – they clicked a download link, it feels like whatever happens, we should download the file, but pressing `esc` or clicking away from the modal, and then having a file download regardless feels a bit weird to me.)

MyfanwyNixon:
Instead of 'Would you mind if we followed up with you' how about 'May we follow up with you'? 

Less 'might' and more 'will definitely happen'.

MyfanwyNixon:
Actually I think it's fine :) Surely people can make that connection.

zarino:
Okeydoke, here’s how it looks with that change made:

Screenshot 2022-04-06 at 09 49 55

@struan I’ve refactored it into its own template, to make including it on the homepage simpler. Only thing left to fix, I think, is downloading only on button press, rather than modal dismiss (see comment above).

zarino (approved):
Looks good to me. The modal javascript seems complicated, but TBF I can’t immediately work out a way to simplify it, so I guess it’s just the right level of complication ;-)

We may find that sending these download events as emails is going to get tiring very quickly, and might want to log them in the database instead. But I guess we’ll cross that bridge when we come to it.

PR #369, opened 2022-04-13
Add filtering by english region and county to CAPE and scorecards

### LINKED ISSUE #364: [CCL] / [CAPE] Allow filtering of local authorities by region and county
Feedback suggests that one other way people expect to compare local authorities is by rough geographical "region", eg: "North West local authorities", or "local authorities in Hampshire".

This feels like a fairly natural way to think about "nearby" / "comparable" local authorities.

And that got me thinking – no reason this should be limited to the Scorecards site. Wouldn’t it be great, in CAPE, if the "Country" filter was replaced with something like:

Screenshot 2022-03-25 at 13 10 18

struan:
Adds region and county filter to scorecards advanced filters.

Adds a region filter to cape the covers country, english region and counties.

Fixes #364


zarino (approved):
Looks good to me. Just check, once you’ve rebased this, that the little "1 active" red badge on the Scorecards "Advanced filters" button works when a region is selected:

Screenshot 2022-04-14 at 11 39 11

(No reason it shouldn’t just happen automatically, but that commit was added _since_ this branch started, so missing from this PR.)

PR #371, opened 2022-04-14
move csv conversion utils to utils

### LINKED ISSUE #214: Move CSV utility conversion methods out of PlanDocument class
Now that the _x_from_text_ methods are used in various places that aren't related to importing plan documents these should really be moved out to a separate class.

struan:
no need for them to be in PlanDocument given they are used for other
things so move them in to the utils module

Fixes #214


ajparsons (approved):
This looks like it all builds ok from scratch, so it's good! 

Meta comment: the pandas way of solving this problem would be to fillnas or apply a function to the column before the loop (but this works fine, and can be read easily).

PR #372, opened 2022-04-14
place caps templates in correct location

### LINKED ISSUE #277: Django tidy up: Namespace caps templates
The templates at `/caps/templates/*` should actually be at `/caps/templates/caps/*`.

struan:
move caps/templates/* -> caps/templates/caps/

avoids chance of namespace overlap

fixes #277


zarino (approved):
Looks good to me! 👍

PR #373, opened 2022-04-18
add a search form on the council detail page

### LINKED ISSUE #219: Search within council plans and similar council plans

struan:
lets people easily search within the council's documents.

Fixes #219

### DIFF
--- a/caps/static/caps/scss/ceuk-card.scss
+++ b/caps/static/caps/scss/ceuk-card.scss
@@ -54,6 +54,10 @@
             background-color: $color-ceuk-grey-100;
         }
     }
+
+    .card-footer {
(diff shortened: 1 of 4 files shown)


struan:
This requires a tiny bit of design attention.

zarino (approved):
Great! I’ve added a couple of styling fixups. Feel free to merge them all into one if you like. Or keep separate. I don’t mind.

Issue #225, opened 2022-01-06
Change from data.climatemergency.uk to a more descriptive URL


ajparsons:
Caching a discussion from slack, where `plans.climatemergency.uk` might better reflect what the site does. 

(Might also be nice to have mysociety.org/cape redirect)

zarino:
Had a quick chat with Isaac at CEUK about this. Given CAPE is increasingly not just about Climate Action Plans any more, it seems counterintuitive to move from the very general `data.` subdomain to `plans.`

CEUK seem happy with `data.` and I figure we should only change if there’s a really clear benefit. (Like if we ever wanted to move it to its own, separate, top level domain.)

ajparsons:
Re-opening because I got annoyed at the URL again. Would really love to at least have a cape.mysociety.org redirect.

struan:
Need to confirm that we are all happy with the URL. @zarino

zarino:
@struan I think the plan was cape.mysociety.org

JenMysoc:
We are fine to go ahead with cape.mysociety.org

struan:
Done.

PR #374, opened 2022-04-18
Add more filtering to CAPE

struan:
Adds filtering by geography.

For #225

### DIFF
--- a/caps/models.py
+++ b/caps/models.py
@@ -16,6 +16,7 @@ from django.conf import settings
 from django.core.files.storage import FileSystemStorage
 from django.db import models
 from django.db.models import Count, Max, Min, Q, Sum
+from django.db.models.expressions import RawSQL
 from django.forms import Select, TextInput
 from django.utils.text import slugify
 from simple_history.models import HistoricalRecords
@@ -160,6 +161,13 @@ class Council(models.Model):
         (None, "All"),
         (True, "Yes"),
         (False, "No"),
+    ]
+
+    PLAN_GEOGRAPHY_CHOICES = [
+        ("urban", "Urban"),
+        ("rural", "Rural"),
+        ("urban-rural-areas", "Urban with rural areas"),
+        ("sparse-rural", "Sparse and rural"),
     ]
     created_at = models.DateField(auto_now_add=True)
     updated_at = models.DateField(auto_now=True)
@@ -844,6 +852,13 @@ class CouncilFilter(django_filters.FilterSet):
         widget=Select(choices=Council.PLAN_FILTER_CHOICES),
     )
 
+    geography = django_filters.ChoiceFilter(
+        method="filter_geography",
+        label="Geography",
+        empty_label="All",
+        choices=Council.PLAN_GEOGRAPHY_CHOICES,
+    )
+
     sort = DefaultSecondarySortFilter(
         secondary="name",
         label="Sort by",
@@ -886,6 +901,17 @@ class CouncilFilter(django_filters.FilterSet):
                 return queryset.filter(**{"country": value})
             except ValueError:
                 return queryset
+
+    def filter_geography(self, queryset, name, value):
+        if value is None:
+            return queryset
+        else:
+            # can't use planscores as importing scoring models results in a circular reference hence raw
+            plans = RawSQL(
+                "select distinct council_id from scoring_planscore ps where ps.year = %s and ps.ruc_cluster = %s",
+                [settings.PLAN_YEAR, value],
+            )
+            return queryset.filter(**{"id__in": plans})
 
     # we do this here as otherwise region_filter_choices is called before
     # migrations have been run which causes problems for tests on github


struan:
@zarino this is running up against issues with the length of the filter options, especially with scrolling. I suspect a bit more of a rethink of how we do this is required?

lucascumsille:
Hi @struan I'm trying to work on this PR, but I get the following error when I change to the 225 branch:

`column caps_council.population does not exist
LINE 1: ..."caps_council"."region", "caps_council"."county", "caps_coun...`

I'm guessing I need to import a new csv file? Thanks

zarino:
Yes, this one contains a fair bit of complexity. But worth it, I think.

If we move the sorting input visually out of the form, then people won’t expect to press the form’s submit button to apply the sorting – so yes, feels like that sorting needs to happen automatically, when the input is changed. And politeness would suggest that, if we’re going to update the table automatically (unexpectedly) when the user changes the sorting dropdown, then we should really do it via ajax, rather than a full page reload. We’ll obviously still need a little submit button next to the sorting dropdown, for non-JavaScript users, which we can hide when JavaScript loads.

It does make me wonder… if we make the sorting work without a page reload, I wonder whether we could make _all_ of the filters work without page reload too. 🤔 

And yes, as @lucascumsille says, since some of the filters will be hidden by default – if one of them is actually _active_, I guess the whole set should be _shown_ instead, to make it clear that you’ve interacted with that part of the page. (Again, this is another thing that will work slightly differently—but I guess we’d get entirely for free—if the filtering happened via ajax.)

struan:
I've pushed a commit that I think solves most of the outstanding issues with this but I have probably committed design crimes in the process so it should probably get another pass over to check it all does the right thing.

I need to think about the showing active filters bit. I'm not intending to address the filtering in place with this because that's going to be quite a bit of work and I'd rather split that out into another ticket.

struan:
I've pushed a commit that sets up some template variables to allow you to see which filters are active and something in the template to use it but I'm really not sure how best to display it.

zarino:
This works really well, thanks @struan!

Definitely needs some design attention, although I don’t have time to do that right now. In case @lucascumsille gets to this before me, my current thoughts are:

- Remove the `col-md-10` from the elements on this page, let them go to full container width.
- Use something like `display: grid; grid-auto-columns: 1fr; grid-auto-flow: column; grid-gap: 25px;` on both "rows" of controls, and remove the `max-width` and `padding-left` on the `form-group`s, to get nice equal width elements, without faff – although will need to replace `houdini-target-flex` as a result (the row will no longer be `display: flex`) and will need to think about how the elements stack on narrow screens.
- Add `text-centered` to the houdini label.
- Remove the `table-responsive` class off the table, and instead _wrap_ the table in a new parent element with the `table-responsive` class applied.

zarino (changes requested):
Looks great 👍 but needs some design attention, as outlined above.

lucascumsille:
Thanks Struan! @zarino I might have some time this week to work one this.

zarino:
I’ve merged @lucascumsille’s front-end styling work from https://github.com/mysociety/caps/pull/415 into this branch now.

```
* c25a237 [CAPE] Fix colour bug in equilateral-triangle mixin <Zarino Zappia>
* aea6ca6 [CAPE] Styling and JavaScript for council list advanced filters <Zarino Zappia>
* dce2adc [CAPE] open up the council list advanced filter panel if active filters <Struan Donald>
* d0253c6 [CAPE] make council table sort in place with javascript <Struan Donald>
* 8fc47bc Minor styling fixes <Lucas>
* c34db15 Change subheader layout <Lucas>
* 57520b6 Added new layout <Lucas>
* 67943de Added filter top bar <Lucas>
* 6aff7e5 Added new filter layout <Lucas>
* ae88cfb Find a council new layout <Lucas>
* cd320d6 Added toggle <Lucas>
* 898fbbc move Northern Irish councils to their own council type <Struan Donald>
* 712d379 add population filtering to CAPE <Struan Donald>
* 14ee0f9 add population data to councils when importing plans <Struan Donald>
* cf2af57 add population field to council model <Struan Donald>
* 7505aed add filtering by geography type to CAPE <Struan Donald>
```

@struan I’m thinking we should probably squash commits `aea6ca6` to `cd320d6` into a single commit – they contain a lot of shuffling back and forth between various front-end layouts, when in reality, I don’t think we’ll care about these in future. Really all that matters was the backend changes were made (your initial 5 commits) then the front-end was built (everything after that). Would you agree?

struan:
Yes, that squashing makes sense. This is going to be as slightly annoying merge as since this started we moved the caps templates to `caps/templates/caps`.

zarino:
@struan I’ve rebased this branch onto the latest `master`. If you’re happy, could you merge and deploy?

zarino (approved):

struan:
Merged in b9fda81

PR #379, opened 2022-04-26
detect postcodes when searching and offer location search

### LINKED ISSUE #107: Detect and handle postcodes entered into search box
Some evidence from analytics that this is happening so we should do the right thing and redirect these to the postcode search.

struan:
People sometimes type postcodes in to the search box and they are most
likely looking for a council and not a postcode in the documents. If we
detect a postcode then offer them a link to lookup the council for that
postcode to make things a bit more friendly

Fixes #107


zarino (approved):
Looks good. I’ve pushed a commit with some styling for this.

PR #382, opened 2022-04-28
add extra document types to plandocument model

### LINKED ISSUE #348: Add new document types
In the scorecards process quite a few new documents were discovered like baseline emissions studies, engagement reports etc. It would be nice to include these with appropriate labels.

Would require updating the spreadsheet with new categories and also the import script and model.

struan:
Allows us to include various other common plan related documents.

Spreadsheet dropdown already updates.

Fixes #348


zarino:
@struan I notice the spreadsheet currently includes some documents with a kind of "Progress Report" – are those currently being imported into CAPE, and if so, what do they display as? Should we add a "Progress report" option to `DOCUMENT_TYPE_CHOICES` in this PR?

struan:
I've added progress report to this.

I should note that if a document has a type in the spreadsheet that isn't included here it'll still be imported and listed on the site as Council Document:

Screenshot 2022-06-13 at 17 06 18

ajparsons (approved):
Looks good! I guess my long term question is at what point this should get refactored as a separate table so it's a data change not a database change?

But 👍

Issue #354, opened 2022-03-17
Handle "empty" document attributes better in front-end


zarino:
We do the most naive thing possible when rendering out the attributes of a document. Which often results in a mess like this:

<img width="408" alt="Screenshot 2022-03-17 at 08 16 49" src="https://user-images.githubusercontent.com/739624/158766265-f8fd4a2a-1f5a-40a6-97cf-b749619e2943.png">

We should use CSS to add the separators between pieces of document metadata, rather than `&middot;` characters, to prevent "empty" list items like this.

zarino:
This will be fixed by my UI changes in #500.

PR #383, opened 2022-04-28
Display at most one of each type of plan document on council page, with a link to reveal more (including deleted plans)

### LINKED ISSUE #145: think about how to archive old plans

struan:
Fixes #145 #354.


struan:
Screenshot 2022-04-28 at 16 51 08

struan:
I wonder though if it makes more sense to have the history per document, as an accordion sort of thing, and then another "previous documents" bit at the bottom? @zarino?

lucascumsille:
Hi @struan Struan
After a meeting with @zarino , we had the following idea:
 The "Climate documents" card would be more organised if we had one section for each document type (Summary, Action plan, Pre-plan, Climate strategy). I could only find four different ones, but let me know if there are more.

In each section, only the most recent version of the document would be displayed, and if there are older documents, we would have "Display older versions of this file". For scannability and because a document file name does not always have "Climate strategy" or a description of what we are opening, I have included a tag next to the most recent file.

Because the date was not the date when the document was updated, we will not keep using it.

Can you change the loop? So we have one loop for each section (document type) displaying the most recent document on top and then the older versions.

I have attached a mock-up

Document history Mockup

Let me know if you have any questions and thanks for your help

struan:
The one slight problem with this is it's possible that there might be more than one document of any particular type. eg:

https://data.climateemergency.uk/councils/bracknell-forest-council/ (2 climate strategies)
https://data.climateemergency.uk/councils/aberdeen-city-council/ (2 pre-plans)

I can gather them by document type and then under each one I can do the older documents if that makes sense?

struan:
Ah, I see. My confusion was I'd viewed this as about the history of individual documents.

zarino:
Yeah, I guess we’re seeing it as a history of documents in each type. Sort of. Will this prove to be a problem?

struan:
I don't think so. I guess we'll find out :)

struan:
Ok, I think I have got a rough version of this in place (houdini is my new hammer) needs quite a bit of styling but hopefully is enough to get on with?

struan:
@zarino @lucascumsille I've pushed a rough version of this including deleted documents with some basic styling.

zarino:
@struan I’ve added a commit that tidies up the styling a bit, and makes deleted documents non-interactive.

My commit also changed the `<summary>` wording from "other documents" to "older documents" … but now, on second thoughts, I’m wondering whether that will always be true? Can we say anything about the relative ages of the documents, in the order they’re sent to the template? Is the first one always the most recently added to our database?

struan:
The order is by recent update so it's not necessarily that they are older.

zarino:
Let’s just go with "other" for now then.

zarino:
@struan can you check what happens with deleted plans, in this PR? And either close or update https://github.com/mysociety/caps/issues/145 with what you find?

caps/templates/caps/council_detail.html
```diff
@@ -140,14 +143,17 @@ <h2 class="mr-auto pr-3 mb-1 mb-sm-0">Climate documents</h2>
...
-                                    {% if plandocument.is_deleted %}
+                                    {% if plandocument.is_deleted and not plandocument.file_exists %}
                                       <span class="mx-2">&middot;</span>
                                       We no longer hold a copy of this document
                                     {% endif %}
                                 </p>
                             </div>
 
                       {% if plandocument.is_deleted %}
+                        {% if not plandocument.file_exists %}
```
zarino, line 154: @struan I _think_ this is meant to be `{% if plandocument.file_exists %}`?

struan: doh! Fixed

zarino:
@struan even after a chat with you I’m not sure I completely understand how this PR will affect the display of actually-deleted versus deleted-but-still-cached plan documents, but if you think it does the right thing, I’m fairly sure the styling is at least all fixed now.

Maybe the next step might be to either try it out locally with come contrived examples of all the possible states, or try it out on a staging site with a copy of the _live_ database? (Is that even possible?)

PR #384, opened 2022-04-28
correctly filter councils on has plan based on document type

### LINKED ISSUE #381: update filters with new definition of what is a plan
If you filter on councils with a plan then you get some that are marked as not having a plan as the filtering still counts councils with any document.

struan:
Fixes #381


zarino (approved):
Seems to make sense! Is this an opportunity to add another test, this time for the filtering code? (Asking as I’m not sure how easy / worthwhile that’d be in Django-land!)

Issue #129, opened 2021-09-20
Improved search and deep linking


ajparsons:
Copying from a notes document:

>One of the user interviews was disappointed that the search couldn’t link to individual pages, and this feels a fixable issue with a little R&D time. Similarly there are some existing tools [for table of contents extraction](https://pypi.org/project/pdf.tocgen/) that paired with deep linking would make PDFs much more useful without needing to fully understand the content. Structured understanding of tables of contents could improve search, and provide a light form of structured content for comparing documents. 

Chrome supports both linking to a specific page of a PDF url, and also linking to specific text fragments on an html page. Could use this to validate if it's useful.

PR #388, opened 2022-05-04
use extract file contents to index documents

struan:
This is a Solr specific thing that relies on Solr's
ExtractingResourceHandler to parse structured files - see
https://django-haystack.readthedocs.io/en/master/rich_content_extraction.html

Means we should be indexing word documents etc as well.

For #129


struan:
supersede by #404

PR #389, opened 2022-05-17
Changed layout "Climate Documents" card

lucascumsille:
I moved the search bar to the top and replaced the label with a placeholder instead. I think the council page is quite busy at the moment, there are a lot of elements that draw attention, such as links and elements with colours, so I thought about economising and getting rid of the label.

I put the google form message at the bottom of the "Climate documents". Same reason as before, I wouldn't use h2 or make the message pop up that much. Give it enough visibility, so users know it is there without making it look like the most important part.

Screenshot 2022-05-17 at 08 30 36

### DIFF
--- a/caps/static/caps/scss/backgrounds.scss
+++ b/caps/static/caps/scss/backgrounds.scss
@@ -5,3 +5,9 @@
 .bg-orange-gradient {
     background-image: linear-gradient(45deg, $color-ceuk-red, $color-ceuk-orange) !important;
 }
+
+@each $name, $color in $theme-colors {
+    .bg-#{$name}-light {
+        background-color: rgba($color, 0.05) !important;
+    }
+}
--- a/caps/static/caps/scss/ceuk-card.scss
+++ b/caps/static/caps/scss/ceuk-card.scss
@@ -97,5 +97,15 @@ $ceuk-card-colors: (
         .card-footer {
             background-color: rgba($color, 0.05);
         }
+
+        .card-body.is-emphasised-section {
+            background-color: rgba($color, 0.05);
+        }
+    }
+}
+
+@each $name, $color in $ceuk-card-colors {
+    .card--#{$name} {
+        background-color: rgba($color, 0.05) !important;
     }
 }
--- a/caps/static/caps/scss/ceuk-card.scss
+++ b/caps/static/caps/scss/ceuk-card.scss
@@ -97,15 +97,5 @@ $ceuk-card-colors: (
         .card-footer {
             background-color: rgba($color, 0.05);
         }
-
-        .card-body.is-emphasised-section {
-            background-color: rgba($color, 0.05);
-        }
-    }
-}
-
-@each $name, $color in $ceuk-card-colors {
-    .card--#{$name} {
-        background-color: rgba($color, 0.05) !important;
     }
 }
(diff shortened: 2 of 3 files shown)


zarino:
@MyfanwyNixon Do you have any thoughts on the wording in the yellow panel (see screenshot above)? It’s only placeholder text right now, and feels a bit wordy to me. What do you think would be the most concise, persuasive wording? Want to make sure a council officer visiting their council’s page would immediately know how to provide an updated document … but ideally it should also be clear to the general public too, in case they’ve come across a newer document and fancy being a good samaritan.

MyfanwyNixon:
@zarino I don't have a massive problem with this text - seems pretty clear to me, but if you wanted to cut it down you could try something like:

**Are these plans out of date?**
[Fill in this form] to let us know where to find the latest plans from this council.

caps/templates/council_detail.html
```diff
@@ -77,6 +77,24 @@ <h2 class="mr-auto pr-3 mb-1 mb-sm-0">Climate documents</h2>
                 <small class="text-muted line-height-1">Last update: {{ last_updated.last_update }}</small>
               {% endif %}
             </div>
+
+            {% if council.plandocument_set.all %}
+            <div class="card-body is-emphasised-section">
+                <form action="{% url 'search_results' %}" data-show-interstitial class="form-row flex-wrap align-items-center">
+                    <div class="col-auto flex-shrink-1 flex-grow-1" style="max-width: 450px">
+                        <input type="text" class="form-control form-control-sm" name="q" id="search-this-councils-documents" value="{{ q|default_if_none:'' }}" placeholder="Search inside these documents">
```
zarino, line 85: This input needs a `<label>`, rather than hiding label text in a `placeholder` attribute. We could make the `label` screenreader-only, though, if you like, to keep the UI minimal.

Wonder whether we should give it `type="search"` too, while we’re at it…

caps/static/caps/scss/ceuk-card.scss
```diff
@@ -97,5 +97,15 @@ $ceuk-card-colors: (
         .card-footer {
             background-color: rgba($color, 0.05);
         }
+
+        .card-body.is-emphasised-section {
```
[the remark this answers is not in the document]
lucascumsille: Sorry that was force of habit. I'll stick to BEM

lucascumsille:
Thanks for the feedback @zarino I have:

-  Changed the copy for the google form message.
- Standardise  how to give a light background-colour to an element
- Added a label visible only for screen readers and the input type to "search"

caps/static/caps/scss/backgrounds.scss
```diff
@@ -5,3 +5,21 @@
 .bg-orange-gradient {
     background-image: linear-gradient(45deg, $color-ceuk-red, $color-ceuk-orange) !important;
 }
+
+$theme-colours: (
```
zarino, line 9: Why define our own `$theme-colours` in this file, when there’s already a similarly named and almost identical `$theme-colors` in `caps/variables.scss`?

I’d just drop this `$theme-colours` declaration, and change the `@each` to loop over `$theme-colors` instead.

zarino (commented):
Looks good. Not sure we need the `$theme-colours` variable though?

caps/static/caps/scss/backgrounds.scss
lucascumsille: Good one, I just added a fixup that fixes the issue.

zarino (approved):
👍

PR #393, opened 2022-05-23
Improved council page accessibility (colour contrast)

lucascumsille:
Hi @zarino
The first commit improves the accessibility issues found on the council page, where teal, cyan, green, and other colours weren't passing the contrast test. I checked the other pages to see if I could find an issue with the changes, but I didn't find any.

The rest of the commits are not about accessibility but scannability on a council page.
Because there are many interactive elements(purple), we have a lot of colour on the page, which is a bit overwhelming and hard to scan. Therefore, I made the following changes to improve scannability:

- DECLARATIONS & PLEDGES card is now grey, so the titles have a neutral colour.
- The cards in the left sidebar are now grey as well. Most of them were purple, which was confusing cause we used purple mostly for interactive elements.
- The ".summary" elements are now grey. My logic is the chevron already shows they are interactive. Also, they are not the most important bit of the page. However, that whole list using purple draws your attention into that section massively, plus being on the left side part of the screen.
- Changed the hero and subheader colour from red to blue. I wanted to tune down that section a bit.

Let me know what you think

Screenshot 2022-05-23 at 15 47 30

Screenshot 2022-05-23 at 15 48 48

Screenshot 2022-05-23 at 15 48 15

### DIFF
--- a/caps/static/caps/scss/variables.scss
+++ b/caps/static/caps/scss/variables.scss
@@ -1,12 +1,12 @@
-$color-ceuk-blue: #00aeee;
+$color-ceuk-blue: #0277D1;
 $color-ceuk-navy: #005cab;
-$color-ceuk-purple: #7360eb;
+$color-ceuk-purple: #4e38de;
 $color-ceuk-pink: #d92b85;
 $color-ceuk-red: #e11d21;
 $color-ceuk-orange: #f29e1a;
 $color-ceuk-yellow: #ffd80b;
-$color-ceuk-green: #28cf6b;
-$color-ceuk-cyan: #54d9e2;
+$color-ceuk-green: #18964a;
+$color-ceuk-cyan: #0f7f8d;
 $color-ceuk-grey-100: #f3f6fb;
 $color-ceuk-grey-200: #ecf0f5;
 $color-ceuk-grey-600: #6c757d;
--- a/caps/templates/caps/council_detail.html
+++ b/caps/templates/caps/council_detail.html
@@ -24,7 +24,7 @@
 <div class="row">
     <div class="col-md-7 col-lg-8 order-md-2">
 
-        <div class="card ceuk-card ceuk-card--orange mb-gutter">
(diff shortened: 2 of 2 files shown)


zarino (commented):
Sorry I’ve left this hanging for so long @lucascumsille!

I think improving WCAG colour contrast compliance is always a good thing. So we should deffo merge that bit of your changes. (One suggestion: `#0277D1` is a more vibrant blue than `#3256d2`, and still achieves 4.6 contrast with white.)

I get why you want to remove the purple colour from the ‘Powers & Responsibilities’ accordion (to de-emphasise that element) but the lack of colour makes these feel less interactive, less inviting to click – to me at least. Really we should solve this emphasis problem by moving the element somewhere else. Maybe just swapping the order of the two columns, as we discussed a few weeks ago?

Beyond that, I feel like the removal of colours from the `ceuk-card`s in `council_detail.html` make the council page slightly _harder_ to navigate, as so much more is now just uniform black/grey. Combined with the now darker blues/purples/cyans/greens, everything feels quite drab and colourless :-/

The colour palette situation on CAPE clearly isn’t ideal. Longer term, my instinct is, rather than darkening everything down, we should really be cutting down the number of colours (we only started off with this huge palette because we were trying to match the CEUK WordPress theme), and using them more judiciously.

And in on the council page in particular, I feel like we’ll be able to get away with less colour / darker colours, if we throw away the current layout, and go for something with a lot more breathing space, and clearer separation / hierarchy between sections. Right now, colours are providing that separation, but once the colours are darkened or removed, we’ll need to use spacing and typography to do that job instead.

What do you think – shall we just improve the colour contrast for now, and leave the other changes to a wider revamp of the council page layout/structure?

lucascumsille:
@zarino I have changed the accent colour for the fist card "Climate & Pledges". Orange will be a tricky colour to achieve the contrast without decreasing saturation and brightness.

zarino (approved):
Thanks for this! :+1:

PR #395, opened 2022-05-24
Migration updates

sagepe:
Some small changes and additions to permit this to run on a VM using a virtualenv.

Note that I've moved the resolution of `PERSONAL_ACCESS_TOKEN` into the local settings file so that this can be handled by our automation more easily, as we can't relay on using the environment. This is more portable between different deployment methods.

Most of the other changes are fairly simple; you might just need to be mindful if running a local development environment using Docker that maps in the repository root if you've also got a local virtualenv (although if you're using compose, you probably won't have one).


struan (approved):
👍

PR #397, opened 2022-05-26
fix key name error in question summary stats object

struan:


ajparsons:
I can see this makes sense because it relates to this line right?

https://github.com/mysociety/caps/blob/e122922d519ed42a29b3e766b1267388a4be64fd/scoring/templates/scoring/question.html#L115

It should fix this?

image

struan:
Yes to both of those.

ajparsons (approved):

struan:
merged in https://github.com/mysociety/caps/commit/63f29e6b0e1e3cbdf9a50cae0657cb863e113fcb

PR #401, opened 2022-05-31
Remove white header bar

ajparsons:
This PR removes the white header bar from CAPE (from when it was pretending to be part of the CEUK website).

image

CEUK is cool (and encouraging) with this, only requirement that they are credited as founders and there is still a prominent link to the checklist.

For the moment, I have just moved the Checklist link to the red bar. In principle this could instead link to a new page (or section of the about page).


zarino (commented):
Wow that’s a lot of links in one menu! Gets a bit crazy on medium-to-narrow width screens. 

I wonder whether maybe we split out "About" and "Data" again (rather than the current "Data & more") and put two short summaries of CEUK and mySociety onto the `/about` page, which can then link to the external sites for more detail. That’d mean we wouldn’t need "About CEUK" and "About mySociety" in the nav (and also means there are two fewer external links in the nav – something that always feels a bit weird).

I’ve also been thinking that maybe the "Browse by feature" page could be working a bit harder for us, and could maybe be morphed into a "Plans" page that could feature lots of different ways to interrogate plans – from searching for text across all plans, to browsing our curated lists of the best plans (the "features"), to jumping off to the 2021 Scorecards. If we did that, then maybe the menu would end up more like: "Councils", "Plans", "Projects" (?), "Data", "About"…? Hmmm.

I will have a go at rationalising all of this, and add more commits here.

zarino:
I’ve removed some items from the nav menu:

Screenshot 2022-07-14 at 17-20-07 Tracking the UK’s journey towards carbon zero

Added introductions to mySociety and CEUK (including links to the Checklist and Scorecards) on the About page:

Screenshot 2022-07-14 at 17-20-29 About

And moved much of the _previous_ `/about` page content onto `/about/data`, since it was all talking about data anyway.

While I was in the area, I’ve reordered the sections on `/about/data` and made the download links stand out more:

Screenshot 2022-07-14 at 17-20-56 Our data

And I’ve added the mySociety logo to the OpenGraph image, alongside the CEUK one, for consistency.

I’m still keen to make changes to the "Browse by feature" page, as mentioned above, but I’ll leave that for another branch.

ajparsons:
This looks really great! The one thing I wonder is if we can make the signposting to the checklist more prominent in the about to make ceuk happy.

Could just double up the #ceuk link. Can add a commit for that if you agree.

image

zarino:
I feel a little bit _eugh_ about including an external link in that _in-page_ navigation menu – like, "these are all sections of this page which you can skip down to—OOP!—except this one, ho ho ho, gotcha!"

I’m tempted to say we deploy what we’ve got, then CEUK can take a look and give us feedback on our call this coming Tuesday.

ajparsons:
Sorry wasn't clear, I wasn't including it as an external link - it was doubling up the link to the CEUK section of the about page (which talks about it - just making it easier to find). 

But happy to deploy and then get their opinions!

PR #402, opened 2022-05-31
[scorecards] add links to raw data

### LINKED ISSUE #362: [CCL] Provide a CSV of all data for journalists/researchers to use
We need to provide the CSV file of all the data on https://councilclimatescorecards.uk, for people to be able to play with the data. And probably need to write some tiny text for the website for this.

Notes from discussion in Google Docs:

> **zarino:** do you think [the CSV files here](https://github.com/mysociety/climate_scorecard_data/tree/main/2021) would be suitable for the people this is aimed at?
>
> **annie:** pretty much, but needs a tiny bit of tidying:
> 1. Round all the numbers to full % (already had examples of people coming up with conflicting totals when using the unrounded data)
> 2. Don't think we need to share [questions.csv](https://github.com/mysociety/climate_scorecard_data/blob/main/2021/questions.csv)?
> 3. And maybe not this [raw_section_marks.csv](https://github.com/mysociety/climate_scorecard_data/blob/main/2021/raw_section_marks.csv) either? Unless it’s useful for data peeps?

Instinct is, what we shared with the Guardian for the launch story is about the right thing to share here.

struan:
It's mostly the link at the bottom of the council table that I'm not sure about design wise.

Fixes #362


scoring/templates/scoring/methodology.html
```diff
@@ -24,6 +24,7 @@ <h3>Disclaimer</h3>
                     council pages. Visit the <a href="https://data.climateemergency.uk/">Climate Action Plan Explorer</a> to see if there is a later version of a council’s plan.</span>
             </div>
             <p><strong>To find out about the history of and methodology behind the Scorecards, please go to </strong> <a class="d-inline-block" href="{% url 'about' %}">About us</a></p>
+            <p><a href="https://mysociety.github.io/climate_scorecard_data_2022/">Download the raw dataset in spreadsheet form</a>.</p>
```
zarino, line 27: FWIW when an entire sentence is a link, we include the period _in_ the link, rather than outside it.

zarino (approved):
@struan I’ve ammended your commit to remove the `.plain` class for links (we can use `d-inline` if we ever need to override links to inline) and then figured, on the homepage, I’d just re-use the `.cta` class that we were _already_ using further up the page, for a bit of consistency.

PR #404, opened 2022-06-08
129 improved document indexing

struan:
Use the built in file extraction for document indexing which allows indexing documents other than PDF.


zarino (approved):
I checked this out and ran `./manage.py rebuild_index`, which appeared to remove my search index and replace it, and I can still find text inside plan documents, so I guess this is working!

PR #406, opened 2022-06-13
Streamline docker compose and image generation

ajparsons:
This PR tidies up the docker and devcontainer config in CAPE to follow a described pattern.

This primarily effects the environments I use (and hope that others will try), but I want to use this as a model that might work for other projects.

The main practical implications are:

- A faster and simplified tests action.
- Configured prebuilds can make a fully functional environment available (with data) in about 2 minutes. [Try it out for this branch](https://github.com/codespaces/new?hide_repo_select=true&ref=container-rationalization&repo=289357345).

@struan : You might be interested/have questions on the change to how the tests action works.
@sagepe: This and the [cached image codespaces pattern](https://github.com/mysociety/codespaces/blob/main/patterns/cached-image-codespaces.md) link are describing my general approach in how docker and codespaces are used in combination. 

# The problems

- Currently there is a `docker-compose.yml` and `docker-compose.dev.yml` which adds a few more dev packages. We only use docker for dev, so these features can be merged back into the main docker-compose  (so `docker-compose up` should always work). 
- The devcontainer is based on an image which a Github Action builds from the `docker-compose.dev.yml` but this is fired manually and is poorly documented. 
- The dev data backup and restore system is similarly undocumented. 
- The tests run by GitHub Action are a third place where configuration is defined for the project. This means the tests do not benefit from the optimisation in docker config, and future updates need to be done in multiple places.

# The solution 

- This repo is now the prototype for the [cached image codespaces pattern](https://github.com/mysociety/codespaces/blob/main/patterns/cached-image-codespaces.md).
  - This creates several devcontainer options for different start-up modes, and better documents the docker image creation (which uses layer caching to speed up rebuilds). 
- A separate GitHub Action automates the generation and back-up of data for the dev database. In combination [with prebuilds](https://docs.github.com/en/codespaces/prebuilding-your-codespaces/configuring-prebuilds) a fresh and populated instance can be created in minutes. 
- The test GitHub Action is replaced by a new action that uses the docker-compose config. It doesn't use the generated image in case there is an update to dependencies in a branch, but the layer caching means this can create the image and complete all tests in 1-2 minutes. 

# Notes

- Before merging, explicit references to 'container-rationalization' in images and actions need to be removed. 
- I've tested the test action by making sure introducing a new breaking error similarly throws an error.
- In principle the separate black linting check could be merged into `script/test`.


script/backup-dev-data
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+#
+# script/quick-setup : Experimental approach to pull in previous test data
```
struan, line 3: this comment doesn't look correct

struan (approved):
I think this looks ok, although as a lot of it is I think VS Code specific I confess to not having tested it, but it looks to make sense.

PR #407, opened 2022-06-13
Fix scorecards rounding error

### LINKED ISSUE #396: [CCL] Minor discrepancies between website and data file
There are a few councils where the rounded total in the data file is off by 1 percent compared to the website.

e.g. Staffordshire County Council is 24% in the data and 25% on the site.

As far as I can see the rounding in the data is right. I think what is happening is the website is rounding twice. On import to 1 dp:

https://github.com/mysociety/caps/blob/master/scoring/management/commands/import_plan_scores.py#L190
```
plan_score.total = round(row["raw_total"] * 100, 1)
plan_score.weighted_total = round(row["weighted_total"] * 100, 1)
```

And then later on the template rounds to 0 DP. This means some things get rounded up to .5, and then to the next percent up. So 0.45 upwards would get rounded to 1.

I think the solution is just to take out that initial rounding, or set that to 5 dp or something. @struan would you agree?

struan:
Fixes #396


scoring/management/commands/import_plan_scores.py
```diff
@@ -186,8 +186,8 @@ def import_overall_scores(self):
                 council=council, year=self.YEAR
             )
 
-            plan_score.total = round(row["raw_total"] * 100, 1)
-            plan_score.weighted_total = round(row["weighted_total"] * 100, 1)
+            plan_score.total = round(row["raw_total"] * 100, 3)
+            plan_score.weighted_total = round(row["weighted_total"] * 100, 2)
```
ajparsons, line 190: Any reason these are different?

ajparsons (approved):
Left a comment out of curiosity, but booted up an instance and it seems to fix the entry I was talking about. I think there's no harm in putting both to 3? But if there's a reason fine.

scoring/management/commands/import_plan_scores.py
struan: The reason is I did this at the end of the day and was tired :| 

So yes, will make both of them 3.

PR #411, opened 2022-06-22
[scorecards] update donate button in page header

### LINKED ISSUE #409: Change donate button for CEUK on Scorecards
From Annie:

"our donate link has now been updated for CE UK.
Could you change the current donate button on the Scorecards site (the crowdfunder) to the one found on this page - https://www.climateemergency.uk/ (this link: https://www.paypal.com/donate?token=ooFAgT1myYlX47dGI06bCoUTffzlA0Sd1JkCKJZDGJ79jsma-Dm0ZGqFib6rxa3w0MCX_9jobx6AEnnm)"

The button is the one in the banner below the header bar

Screenshot 2022-06-15 at 14 55 07

The link it should go to is: https://www.paypal.com/donate?token=ooFAgT1myYlX47dGI06bCoUTffzlA0Sd1JkCKJZDGJ79jsma-Dm0ZGqFib6rxa3w0MCX_9jobx6AEnnm

struan:
I've used the paypal button from the CEUK page as the link provided didn't seem to work and also I suspect paypal requires you to use their button.

Screenshot 2022-06-22 at 10 34 44

What I can't work out is how to disable the hover effect of the background colour showing through. I've set the transition to none but no joy. Possibly @lucascumsille can fix this.

Fixes #409

### DIFF
--- a/scoring/static/scoring/scss/input.scss
+++ b/scoring/static/scoring/scss/input.scss
@@ -48,3 +48,10 @@ input {
     background-repeat: no-repeat;
     background-position: right 10px center;
 }
+
+// CEUK donate button
+input#ceuk-donate {
+    background: none;
+    color: inherit;
+    border: none;
+}
--- a/scoring/templates/scoring/base.html
+++ b/scoring/templates/scoring/base.html
@@ -69,8 +69,8 @@
         </div>
     </nav>
     <nav class="subheader bg-blue-gradient">
-        <span>Please help Climate Emergency UK by supporting our Crowdfunder. We rely on donations to remain independent.</span>
-        <a class="btn donation ml-2" href="https://www.crowdfunder.co.uk/p/council-climate-scorecards">Donate now</a>
+        <span>Please help support Climate Emergency UK. We rely on donations to remain independent.</span>
+        <form action="https://www.paypal.com/donate" method="post" target="_top"><input type="hidden" name="hosted_button_id" value="JB823YSUV5X9S"><input id="ceuk-donate" type="image" src="//www.paypalobjects.com/en_GB/i/btn/btn_donate_LG.gif" name="submit" title="PayPal - The safer, easier way to pay online!" alt="Donate with PayPal" border="0"><img alt="" src="//www.paypal.com/en_GB/i/scr/pixel.gif" width="1" height="1" border="0"></form>
     </nav>
 
 {% block content %}{% endblock %}


scoring/static/scoring/scss/input.scss
```diff
@@ -48,3 +48,10 @@ input {
     background-repeat: no-repeat;
     background-position: right 10px center;
 }
+
+// CEUK donate button
+input#ceuk-donate {
+    -moz-transition: none;
+    -webkit-transition: none;
+    transition: none;
```
zarino, line 56: Rather than these `transition` lines, you probably want something like this:

```scss
#ceuk-donate {
background: none;
color: inherit;
border: none;
}
```

(Removes the background and border, and thanks to the ultra-high specificity of the ID selector, this trumps even the hover effects on the more generic `input:hover` … and also sets the `color` to inherit the default dark text colour of the parent so that, in the unlikely situation that the button image fails to load (it’s a third-party resource – could happen!), the fallback alt text will be a legible black on pale green, rather than white on pale green.)

scoring/templates/scoring/base.html
```diff
@@ -69,8 +69,8 @@
         </div>
     </nav>
     <nav class="subheader bg-blue-gradient">
-        <span>Please help Climate Emergency UK by supporting our Crowdfunder. We rely on donations to remain independent.</span>
-        <a class="btn donation ml-2" href="https://www.crowdfunder.co.uk/p/council-climate-scorecards">Donate now</a>
+        <span>Please help support Climate Emergency UK. We rely on donations to remain independent.</span>
+        <form action="https://www.paypal.com/donate" method="post" target="_top"><input type="hidden" name="hosted_button_id" value="JB823YSUV5X9S"><input id="ceuk-donate" type="image" src="//www.paypalobjects.com/en_GB/i/btn/btn_donate_LG.gif" name="submit" title="PayPal - The safer, easier way to pay online!" alt="Donate with PayPal button" border="0"><img alt="" src="//www.paypal.com/en_GB/i/scr/pixel.gif" width="1" height="1" border="0"></form>
```
zarino, line 73: Such a small thing, but I _think_ the `alt` text of the `<input type="image">` here is what will get read out in assistive devices (it is, at least, in VoiceOver on Mac) and as a result, users will hear something like:

> "Donate with PayPal button"

or worse (as is the case in VoiceOver):

> "Donate with PayPal button, button"

I’d drop the "button" from the `alt` text – it makes sense without it.

zarino (changes requested):
A couple of small fixes.

scoring/templates/scoring/base.html
lucascumsille: No worries @zarino I'll take a look on this sprint.

struan: I've pushed a fixup for both these things.

zarino (approved):

PR #414, opened 2022-06-28
[CAPE] fix modal submit handlers

struan:
The submit handler for the interstitial was no longer setting the never
show again property in local storage, but the download modal was so swap
these around.

### DIFF
--- a/caps/static/caps/js/main.js
+++ b/caps/static/caps/js/main.js
@@ -180,7 +180,7 @@ $('#feedback-email').on('keyup change', function(){
     }
 })
 
-$('form[data-ajax-feedback-submit]').on('submit', function(e){
+$('form[data-ajax-submit]').on('submit', function(e){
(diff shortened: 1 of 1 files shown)


zarino (approved):
Yikes, how did this slip past us!

I wonder whether this ~~(showing our feedback popup on every pageload for newish visitors over the last few weeks)~~ is a lesson to finally make us write some integration tests for the JavaScripty bits of the site? Cypress tests for the feedback and data download modals would have caught this typo.

Edit: Turns out it wasn’t quite as bad as this – the `skipped-interstitial` logic still worked, so people who _skipped_ the form wouldn’t see it again, but people who _submitted_ it would. I think…

struan:
It won't have been doing that as if they close it then the "do not show for a week" thing still works so I guess they'd only have seen it twice. Still not great.

PR #415, opened 2022-07-04
[CAPE] Additional filtering design

lucascumsille:
This PR has four commits:
The first one tackles all the observations made by @zarino [in here](https://github.com/mysociety/caps/pull/374#issuecomment-1168520738), such as:

- Full width for the filter and table
- Grid display for filters
- Responsive table
- Houdini label is centred

Everything is working fine till that commit. However, I tried to give a visible focus state for the advanced filter button, but it didn't work.

To make it more accessible I added the next 2 commits that changed the Houdini label and input into a button element. I realised that replacing the advanced filter checkbox with a button broke the feature that kept the advanced filter open when one of them was active. So I added a function to detect if any options on the select element were different from "All". If so, keep the div with the advanced filter open. However, the sort by is not working on these two commits, and I didn't realise why it is happening.

The last commit is kind of unrelated, but adds the "btn--open-div" as a proper component so that we can reuse it when we need it.


lucascumsille:
If we end up using the last commit I could replace the button on score-card table on a council page, that way we could get rid of the specific code on `main.js` just for that element.

caps/static/caps/js/main.js
```diff
@@ -312,6 +312,32 @@ if ($selectedArray.every( (val, i, arr) => val === arr[0])) {
...
+        $btn.removeClass('open');
+        $target.removeClass('open');
+        $btn.text($defaultButtonText);
+    } else {
+        $btn.addClass('open');
+        $target.addClass('open');
+        $btn.text($closeButtonText);
+    }
+});
+// END OPEN DIV BUTTON
```
struan, line 340: I'm not sure I understand what this is for?

caps/static/caps/js/main.js
```diff
@@ -312,6 +312,32 @@ if ($selectedArray.every( (val, i, arr) => val === arr[0])) {
...
+    Filter by authority type and others...
+</button>
+<div class="open-target"></div>
+
+In a case by case, the open-target should have the neccesary display property
+By default it will be block
+*/
+$('.btn--open-div').each(function(){
+    var $btn = $(this);
+    var $target = $btn.closest.next('.open-target');
```
struan, line 327: missing () at the end of `closest`

caps/static/caps/js/main.js
struan: Also, it feels like this is in contrast to the block that opens the div if there's a filter selected.

caps/static/caps/js/main.js
```diff
@@ -271,27 +271,48 @@ $('.scorecard-table').each(function(){
...
+// This part check if there is any advanced filter active
+var $selectedArray = [];
+$(".open-target select").each(function() {
+    var $selectElement = $(this);
+    $selectedArray.push($selectElement.find(":selected").text());
 });
 
+if ($selectedArray.every( (val, i, arr) => val === arr[0])) {
+    console.log( $selectedArray)
+    $btn.removeClass('open');
```
struan, line 304: debug here and below

struan (changes requested):
Looks good but there's some JS I don't really understand/think is needed.

lucascumsille:
Thanks @struan I added some fixups amending the issues you found.

caps/static/caps/js/main.js
lucascumsille: You are right, those functions were conflicting with each other. So I ended up renaming the variables for the `#button-advanced-filter`. This one also uses different button text, so the code for `.btn--open-div` with the "Show more" and "Show less" text won't suffice. 

[The [CAPE] Added button to open div](https://github.com/mysociety/caps/pull/415/commits/a9bb96280d6a2676f7196646ee3d73597224c23d) was included so we can have a button we can reuse, not only for the .scorecard-table, but anywhere on the website.

struan (approved):
this looks good.

zarino (approved):
In reviewing this, I was struck by how overcomplicated all of the markup, CSS, and JavaScript had got – in both this branch and [225-additional-filtering](https://github.com/mysociety/caps/tree/225-additional-filtering). It’s nobody’s fault, just the result of an incredibly long, drawn-out ticket that had passed between multiple people many times, with changing goalposts all the way along.

I’ve had a go at simplifying the markup, CSS, and JavaScript, to make it easier to maintain in future.

One thing I couldn’t work out how to do, was to add a `form` attribute to the sorting field (inside `<div id="sort-by-wrapper">`) so that we could drop the sort field _inside_ the form, and just use that one _outside_ the form instead. I think it would require creating another subclass of `DefaultSecondarySortFilter`, which accepts a `form` parameter from the template. I’ve added `<!-- TODO -->` comments to the template which hopefully explain what I mean.

@lucascumsille can you check out this branch, pull, and check you’re happy with how it looks? Then we’ll get this squashed into a single commit, and merged into [225-additional-filtering](https://github.com/mysociety/caps/tree/225-additional-filtering).

lucascumsille:
@zarino  Looks good, thanks for the feedback.
I just added a commit that hides the advanced filter button when javascript is disabled, but don't use the commit if the extra lines of code are not really worth it.

struan:
This does mean that at the moment you can't sort without JS enabled which is not ideal?

I realise the `form=` thing is meant to fix that but having had a look it's not obviously trivial and I think getting this live and maybe opening a second ticket to fix that is the best thing to do.

PR #421, opened 2022-08-31
Add Github Action to push changes to mirror

ajparsons:
This PR addresses the issue that people working in codespaces (ok, *me*) can't easily push back to `git.mysociety.org` without syncing on the local computer. `git.mysociety.org` is our primary git repository, and other developers not unreasonably expect the code for PRs to be in there. 

[Talking it over with Sam](https://github.com/mysociety/sysadmin/issues/1655) - the simplest thing to do is to add a GitHub Action that pushes changes made to github back to git.mysociety.org. This PR adds that action, and if the linting check succeeds will push to git.mysociety.org.

As they are both trying to push to each other, a push made to either place should end up in both:

```mermaid
graph TD

github[GitHub]
mysoc[git.mysociety.org]
mysoc -->|Post push hook| github
github -->|GitHub Action| mysoc

```

As git.mysociety.org is still the primary, it will take precedence when there's a conflict. Changes that involve force pushes will need to be made to git.mysociety.org. Developers working in the current expected way should have to make no changes. 

This does mean that Github's PR merge tools can be used in browser, so that might be a plus. 

Merging this PR resolves the last major issue in using codespaces for development, and means I have an easier time of making changes, but also more generally I can use CAPS as an example repo for any other projects who wanted to emulate it.


zarino (approved):
If 1) @sagepe’s happy from a security PoV, 2) this works for you, and 3) it won’t cause any issues for the rest of us interacting with this repo via the git.mysociety.org mirror, then it’s fine by me 👍

.github/workflows/mirror.yml
```diff
@@ -0,0 +1,32 @@
...
+        fetch-depth: '0'
+        
+    - name: Push branch to git.mysociety.org
+      id: push_to_mirror
+      uses: mysociety/action-git-pusher@v1.1.0
+      with:
+        git_ssh_key: ${{ secrets.PUBLICCVS_GIT_KEY }}
+        ssh_known_hosts: ${{ secrets.GIT_KNOWN_HOSTS }}
+        tag: ${{ github.ref_name }} 
+        remote: 'ssh://gh-public@git.mysociety.org/data/git/public/caps.git'
```
zarino, line 32: Tiny thing, but missing a newline here.

ajparsons: Fixed!

PR #427, opened 2022-09-06
2020 emissions data

ajparsons:
This PR moves the data processing of BEIS data out to https://mysociety.github.io/la-emissions-data/ and updates for the 2020 BEIS data.

This data also moves Agriculture out to a top-level category. I have updated the chart and page to display this.


struan (approved):
looks good!

PR #428, opened 2022-09-06
Switch source of similarity data to new source

ajparsons:
This switches the source of the similarity comparisons to https://mysociety.github.io/local-authority-similarity/

Because some of the outlying sources have changed, this does adjust some of the scores - but seems sensible enough based on a spot check of a few.


caps/management/commands/common.py
```diff
@@ -0,0 +1,27 @@
+import pandas as pd
```
struan, line 1: presumably this can get dropped if #427 is merged first?

struan: also, given we're likely to be using this in a) data things, b) other climate projects I wonder if we should make a proper package with these functions 
we can add as a requirement?

struan (approved):
looks good!

caps/management/commands/common.py
ajparsons: Yes, that's a good idea. I'll do that and update both PRs.

ajparsons: @struan : both PRs now use the external `mysoc-dataset` package. Because of how I set the minimum requirements, I've had to bump the pandas from 1.3.4 to 1.3.5.  If you give me a thumbs up I can do a big merge of all these PRs.

struan (approved):
👍

PR #439, opened 2022-10-26
Switch to new style cookie disable settings

ajparsons:
Using the wrong style for cookie consent in GA-4.

See: https://stackoverflow.com/questions/67078898/how-to-disable-cookies-in-ga4

This switches to the GA4 style.

### DIFF
--- a/caps/templates/caps/base.html
+++ b/caps/templates/caps/base.html
@@ -35,9 +35,16 @@
             function gtag(){dataLayer.push(arguments);}
             gtag('js', new Date());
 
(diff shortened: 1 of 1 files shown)


struan (changes requested):
This does still seem to set some google cookies but I assume while it does that it turns off tracking at the server side?

Do we also need to do this for scorecards?

struan (approved):
Hmm, not sure if I was looking at the wrong thing but seems to not set cookies :|

PR #452, opened 2022-11-16
Methodology2023 front end

lucascumsille:
Hey Zarino 
This would be the draft, I think I made the accordion + Checkbox work. But I still need the Type of council + council search to work.

Besides the classes that we are currently working I had to add a couple of extras for the "This question might not apply to you"

`.js-single-tier,
.js-district,
.js-county,
.js-northern-ireland,
.js-scotland,
.js-wales`

You can see them in the `methodology.scss` at the end.

I'll start bringing the copy to the file. Let me know if you have any questions.


lucascumsille:
Hey @zarino 
I have pushed a rebased version so we don't have many --fixup commits.
The last commit improves accessibility, after working on FMS I learned that having the same state for hover and focus, It is not the best way to go, so I made both status fairly different. This should affect, buttons, inputs, links and navbar. Let me know if you want to include it in this PR or we can work further and include it on its own one.

Let me know if you have any feedback.

scoring/static/scoring/scss/typo.scss
```diff
@@ -1,4 +1,4 @@
-body {
+body, .body {
```
zarino, line 1: Where does a class of "body" get used? I can’t find it anywhere.

zarino (changes requested):
As mentioned in Slack, apologies in advance for the length of this list. I started noting a few points, then soon realised this was expanding beyond just "a few things to fix" and instead could probably do with some discussion.

I also know we’re working with a short timescale and a couple of competing priorities right now, but I don’t think that’s an excuse for us to drop our usual attention to detail :-)

Some immediate quick fixes, in the order you’d come across them on the page:

- [x] Why is the text between the "Summary" and "Question Criteria" headings inside its own row/col? It’s adding extra padding for no reason. I’d remove it.
- [x] Same with `#section-list-of-questions` and `#section-developing-the-scoring-system` having `.row` classes. No need, that I can see?
- [x] Sidebar nav looks weird touching the bottom of the header/hero. Move it down.
- [x] "pdf" should always be in capital letters: "PDF"
- [x] No need for "Section Weightings" table to be so wide.
- [x] Typo – "Sections Weighting" link in sidebar nav should read "Section Weightings"
- [x] Capitalisations in headings and sidebar nav are all over the place. We usually stick to sentence case, to avoid this. I’d do that here.
- [x] UX point – the list of questions is much too far from the controls that show/hide it – when you make a selection from the form controls, it’s not clear that anything has changed on the page. I _really_ think we should either move the "Section Weightings" table down below the "List of questions", or we should move the form _after_ the "Section Weightings" table
- [ ] UX point – the hit target for the "+" icon in the question accordion is tiny. Why not make the entire accordion headers clickable?
- [ ] Speaking of which, a minor design point – the unequal padding around the `.accordion-heading` `<h5>`s looks sloppy to me. Either reduce the vertical spacing (which is being increased by the `.btn-open-accordion`) or add more `padding-left`.
- [ ] Typo – the "Show all text for the criteria and clarification as default." label should probably end "by default", and no need for a full stop either, I‘d say.
- [x] CSS bug – The "Show all text…" checkbox and label don’t wrap nicely on narrow screens.
- [ ] UX point – We call them "criteria and clarifications" in the checkbox label, but they’re called "question details" in the buttons below each question. We should pick one and use it throughout.
- [ ] UX point – "20% section weight of total score" doesn’t make sense to me. I feel like "This section will contribute 20% to a [council type]’s overall score" would make more sense.
- [ ] "This question might not apply to your council" warning should only appear for Single Tier councils, since Single Tier is the only category that includes councils from multiple countries (District and County are 100% English, and Northern Ireland is its own country too).
- [ ] JavaScript bug – Sometimes closing a question accordion section won’t remove the `.active` class from the `.accordion-heading` or `.accordion-data`, but _does_ give a `style="height: 0px;"` to `.accordion-data`. What is going on there?
- [ ] CSS bug – weird line wrapping for the "You can see all the trial FOI requests on WhatDoTheyKnow." link in the "Trial Marking" section. Why does it have `d-inline-block`?
- [ ] The vertical spacing for the bullet-pointed list of names under "Thank You" is very large. This is made worse when some items wrap onto two lines, while the item next to them doesn’t. Removing the `.thanks-grid` class and replacing it with something like `column-count: 2; column-gap: 40px;` is a much simpler solution (40px to match the 40px default padding-left that my browser appears to be giving to the list element).
- [x] Can’t all of the `.names-wall` elements go into a single `.names-wall-row`? Looks weird having only two or three names per row, when there’s so much space for them to flow into. This would also fix a CSS bug on narrow screens, where there’s a lack of horizontal spacing between `.names-wall-row-item` elements because the `.names-wall-row` shrinks.
- [ ] Minor typo – "hamburguer-black.svg" should probably be "hamburger-black.svg" ;-)

I’ll try to make some of these fixes today, and check them off as I make them, so that we’re not duplicating work.

More subjective points we might want to discuss:

- "Check out our process" in the header/hero is a waste of space. We should either replace this with an actual summary of the page, or remove it. Honestly, at this point, I’m wondering whether the page even needs that massive grey header/hero at all, if all it’s housing is the page title.
- UX point – on medium-width screens, the sidebar nav toggle (a dark blue circle with three lines) feels too easy to miss. Adding a text label inside the button, like "Jump to a section" might help?
- I’m also suspicious that we don’t actually need to hide the sidebar nav so soon. Could keep it visible all the way down to something like the 992px breakpoint at least.
- ~~Too much space inside end of yellow "Combined Authorities" box – paragraph needs a `mb-0` class. But, honestly, it feels like this box is way too prominent for what it is. I wonder whether the same content could be restyled as a `<h3>` and some `<p>`s, and put at the end of the "Question Criteria" section?~~
- ~~The "Currently displaying" grey box takes up far too much space given it’s useless most of the time. The "small number of questions might not apply" wording could be a paragraph of its own, right above the question accordion, and would only be shown when Single Tier is selected.~~
- Hovering over "Question Weighting" headings below each question reveals a popup reading "Look at me I'm a nice hidden message". Do we still need this popup?

And one CSS bug (?) I couldn’t get to the bottom of:

- Something weird is going on when I interact with the page, but I can’t consistently replicate it. Sometimes when I show/hide the questions, fold/unfold the accordions, or whatever, the entire `#section-list-of-questions` element will appear to get narrower. I wonder whether it’s a CSS overflow/padding/margin interaction, related to the smooth "unfolding" animation? Firefox 107. When I open the developer tools to investigate, it fixes itself. Grrrr.

lucascumsille:
Thanks for the input and all the work @zarino. I appreciate it the time. 
I'll be working now on the rest of the things that hasn't been covered.

lucascumsille:
@zarino 
I couldn't replicate this problem. Might have been fixed with one of my latest commit.

JavaScript bug – Sometimes closing a question accordion section won’t remove the .active class from the .accordion-heading or .accordion-data, but does give a style="height: 0px;" to .accordion-data. What is going on there?

scoring/static/scoring/scss/typo.scss
lucascumsille: This is been used on the `.hidden-message ` on the utilities file.

lucascumsille:
@zarino 

> Hovering over "Question Weighting" headings below each question reveals a popup reading "Look at me I'm a nice hidden message". Do we still need this popup?

I just asked CEUK what they want in that message.

lucascumsille:
> "Check out our process" in the header/hero is a waste of space. We should either replace this with an actual summary of the page, or remove it. Honestly, at this point, I’m wondering whether the page even needs that massive grey header/hero at all, if all it’s housing is the page title.

That's a placeholder, I thought CEUK would change it. I just asked them if they want to add something relevant there. Otherwise we can just get rid of it.

zarino (approved):

PR #461, opened 2022-12-02
[CAPE] Refactor site URL into context mixin

### LINKED ISSUE #146: add site url setting to allow absolute URLs in templates
Mostly for use in open graph tags, eg:

https://github.com/mysociety/caps/pull/144/commits/a206dec5187d316a2fb41ddcc7662c86d342d150#diff-680816bd777160aa812e28cdf2f2df2f5f7a957496b988a5cc6e5dbc95cd83a7R18

zarino:
Enables us to keep the site URL out of templates, making it easier to keep absolute links up to date if we ever change URL (which we’re likely to do as part of https://github.com/mysociety/caps/issues/225).

It was tempting to use the domain from `request.META.HTTP_HOST` but this is provided by the client, not the server, so ideally shouldn’t be trusted.

Eventually, I guess, we might want `SiteUrlMixin` to pull the URL out of a server environment variable, so it could be set by a provisioning script, or whatever. But hard-coding it in `mixins.py` seemed a good first step.

Fixes #146.

### DIFF
--- a/caps/templates/caps/base.html
+++ b/caps/templates/caps/base.html
@@ -1,5 +1,6 @@
 {% load static %}
 {% load pipeline %}
+{% load hosts %}
 
 <!doctype html>
 <html lang="en">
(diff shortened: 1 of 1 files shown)


zarino:
@struan thanks for spotting a much simpler way to do this! I’ve rebased and amended the commit message, and will merge now.

Issue #462, opened 2022-12-02
Missing mobile menu icons on CAPE and Scorecards – spaces being stripped from inline SVG data URIs, in CSS


zarino:
I noticed a while ago, the mobile menu icon was missing from the top right corner of the Scorecards site:

![Screenshot 2022-12-02 at 13-10-17 Council Climate Plan Scorecards](https://user-images.githubusercontent.com/739624/205300506-ea2506b8-39cc-4bd9-b523-8255ef5697fd.png)

I assumed it was either a network error at my end, or a temporary caching thing.

Today I noticed it was missing on CAPE too:

![Screenshot 2022-12-02 at 13-10-10 Tracking the UK’s journey towards carbon zero](https://user-images.githubusercontent.com/739624/205300589-632d2bf8-360a-47c0-ba45-23df67944643.png)

And on thinking closer about it, I realised this menu icon (three horizontal lines) is implemented as an SVG, URL-encoded into a Data URI, inside a `background-image:` rule in the site’s CSS … so it _couldn’t_ be an issue with a missing image file or whatever.

Inspecting the CSS sent to my browser, I can see the rule:

```css
.navbar-dark .navbar-toggler-icon {
 background-image:url("data:image/svg+xml,%3csvgxmlns='http://www.w3.org/2000/svg'width='30'height='30'viewBox='003030'%3e%3cpathstroke='%23fff'stroke-linecap='round'stroke-miterlimit='10'stroke-width='2'd='M47h22M415h22M423h22'/%3e%3c/svg%3e")
}
```

Compare to the same rule, when I visit the site running _locally_ in a Docker container:

```css
.navbar-dark .navbar-toggler-icon {
  background-image: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' width='30' height='30' viewBox='0 0 30 30'%3e%3cpath stroke='%23fff' stroke-linecap='round' stroke-miterlimit='10' stroke-width='2' d='M4 7h22M4 15h22M4 23h22'/%3e%3c/svg%3e"); }
```

Aside from the handling of the final `}` (which is interesting – it suggests the formatting/minification settings are _different_ in the docker and live environments) the big problem, the one causing the missing lines in the screenshots above, is that all of the spaces have been stripped from the Data URI, resulting in an SVG that starts with `<svgxmlns`. Since this isn’t a valid SVG start tag, the browser ignores the image.

@sagepe any idea what might be stripping the spaces out of this Data URI?

[...]

PR #463, opened 2022-12-02
`variant: datauri` for Django Pipeline

zarino:
Bit hard to test this without deploying it to the live site (as Django Pipeline isn’t used in the local dev environment), but maybe this might fix the SVG/Data URI issue in #462.


zarino:
Turns out this did not fix it.

PR #464, opened 2022-12-06
[Scorecards] Methodology2023

lucascumsille:
Hi @zarino 
Small PR with some changes:
-Copy change here:  Gov & Finance - question 4
-Added a download button for question spreadsheet on excel format.

Let me know if there is any feedback

### DIFF
--- a/scoring/static/scoring/scss/buttons.scss
+++ b/scoring/static/scoring/scss/buttons.scss
@@ -121,7 +121,6 @@ a.cta {
     letter-spacing: 0.03em;
     padding: 7px;
     text-decoration: none;
-    white-space: nowrap;
 
     &:focus {
         background-color:$color-scorecard-yellow !important;
--- a/scoring/static/scoring/scss/buttons.scss
+++ b/scoring/static/scoring/scss/buttons.scss
@@ -162,7 +162,7 @@ a.cta {
     border: 1px solid $color-scorecard-green-d1;
 
     &:hover {
-        color: $color-scorecard-green-d1;
+        color: $black;
         background-color: $color-scorecard-green-l1;
         border: 1px solid $green;
     }
--- a/scoring/static/scoring/scss/buttons.scss
+++ b/scoring/static/scoring/scss/buttons.scss
@@ -129,6 +129,10 @@ a.cta {
         text-decoration: underline;
         text-underline-offset: 3px;
     }
+
+    &:visited {
+        color: inherit;
+    }
 }
 
 // Color button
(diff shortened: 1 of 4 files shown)


zarino (changes requested):
Sorry to cause you extra work, but I’ve just noticed all of these buttons (including the one at the top of the page, linking to the 2022 methodology) cause a horizontal overflow on narrow screens, because `.btn { white-space: nowrap; }`.

Also, again not something you’ve changed in this PR, but close by – the wording of "Click here to see the draft methodology explanation text in pdf format." seems odd, given it’s not actually a clickable element.

I wonder whether maybe something like this would solve both problems:

```html
<div class="alert">
  <h2>Get this information in alternative formats</h2>
  <p>The explanatory content of this page is available as a PDF:</p>
  <p>
    <a class="btn" href="#">Explanatory content (PDF)</a>
  </p>
  <p>The questions and criteria are available in two additional formats:</p>
  <p>
    <a class="btn" href="#">Questions (PDF)</a>
    <a class="btn" href="#">Questions (Spreadsheet)</a>
  </p>
</div>
```

lucascumsille:
> Sorry to cause you extra work, but I’ve just noticed all of these buttons (including the one at the top of the page, linking to the 2022 methodology) cause a horizontal overflow on narrow screens, because `.btn { white-space: nowrap; }`.
> 
> Also, again not something you’ve changed in this PR, but close by – the wording of "Click here to see the draft methodology explanation text in pdf format." seems odd, given it’s not actually a clickable element.
> 
> I wonder whether maybe something like this would solve both problems:
> 
> ```
> <div class="alert">
>   <h2>Get this information in alternative formats</h2>
>   <p>The explanatory content of this page is available as a PDF:</p>
>   <p>
>     <a class="btn" href="#">Explanatory content (PDF)</a>
>   </p>
>   <p>The questions and criteria are available in two additional formats:</p>
>   <p>
>     <a class="btn" href="#">Questions (PDF)</a>
>     <a class="btn" href="#">Questions (Spreadsheet)</a>
>   </p>
> </div>
> ```

Thanks for that @zarino.

- So I added a fix that gets rid of the extra `white-space` rule. Checking on the website I couldn't see anywhere where getting rid of this cause a problem.
- I spotted an issue on the homepage, presumably when we got rid go the display table for the a.cta class. I just send a commit for that.

Screenshot 2022-12-06 at 11 07 21

- I made an improvement on the` btn-green` on hover wasn't really accessible.

Let me know if there is anything else that needs changing =)

PR #466, opened 2022-12-08
[Scorecards] 3rd round of feedback for methodology

lucascumsille:
@zarino Not sure if you want to wait for extra feedback or we just should move ahead and deploy.

Changes so far:
- Updated copy 90 questions to 89 questions. (I left a message to CEUK, because I could only find one coincidence)


zarino (approved):
:shipit:

PR #482, opened 2023-02-16
[CAPE] Added 20th anniversary badge

lucascumsille:
Hi Zarino
I just added the badge to CAPE. Previews here:

Screenshot 2023-02-16 at 11 12 57
Screenshot 2023-02-16 at 11 12 53
Screenshot 2023-02-16 at 11 12 29

For the image src attribute I had to add a load static at the top of the header file. Not sure if that has performance issue or if there is a better way, like using background-image instead.


zarino (approved):
Spotted a weird little bug where, when the window was between 400–403px wide (!!), the menu button broke onto a second line:

Screen Shot 2023-02-16 at 15 46 49

Anyway, have added a fixup commit for that. If you’re happy, feel free to squash the commits into one.

I guess we want to hold off on merging and deploying this, until the mysociety.org webpage is finished?

PR #484, opened 2023-02-20
update to fetching latest combined authority mapping data

### LINKED ISSUE #465: Import/linking of combined authorities is not working
I think this is because the external data has moved/disappeared. I also note that we are using out of date information in that we are using 2019 data and there is 2021 data available. The 2021 data is in an excel sheet so will need a bit of work for fix.

I think also that the related data repository we get this from will also need an update to get this all to match.

struan:
now using the december 2022 data on local authority -> combined authority mapping

Fixes #465

### DIFF
--- a/caps/management/commands/link_combined_authorities.py
+++ b/caps/management/commands/link_combined_authorities.py
@@ -1,16 +1,13 @@
 from os.path import join
 
-import requests
-
-from django.core.management.base import BaseCommand, CommandError
-from django.conf import settings
-
 import pandas as pd
-
+import requests
 from caps.models import Council
+from django.conf import settings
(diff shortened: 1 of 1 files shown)


ajparsons:
Obvs if it works it's fine - but why not use the https://pages.mysociety.org/uk_local_authority_names_and_codes/datasets/uk_la_past_current/latest lookup? there's a combined-authority column that can work as a map, and we'll keep it up to date in the long run.

struan:
Mostly as I wanted the quickest possible fix which this was. I will open a ticket to use our data as a follow up task.

ajparsons (approved):
Looks good - all GSS codes in this lookup are present in our gss list for plans (so we have none out of date, which is a risk).

Issue #480, opened 2023-02-08
[Scorecards] Add Combined Authority information to Methodology page


lucascumsille:
A few changes to the [2023 draft methodology page](https://councilclimatescorecards.uk/methodology/)

- Addition of Combined Authority info, including link to blog post about CA powers and section weightings (note: CAs have _different sections_ to the other council types)
- Addition of "all the questions" spreadsheet link
- Other minor fixes (typos and a few clarifications)

Final content should be ready for us to use by 20th/21st February, and we’ll need to have this live by 27th Feb.

Full details: https://mysociety.slack.com/archives/CHL40EKB9/p1675846802633599

lucascumsille:
Branch: scorecards-ca-methodology

lucascumsille:
Branch: scorecards-ca-tidy-up

JenMysoc:
Can we close this?

lucascumsille:
Hi @JenMysoc yep it's all good now.

PR #486, opened 2023-02-23
[Scorecards] Combined authorities  methodology 2023

lucascumsille:
This PR includes:

- The latest [web copy](https://docs.google.com/document/d/1wrj23W_Wa3dNVPebUlqowdpb4RQJy2DWAnCJOMpGFcw/edit)
-  The [question list for CA](https://docs.google.com/spreadsheets/d/1cc0PvkSaEngS-9Xti1lrUMftLPDJm59_Y4WL5zW6--A/edit#gid=1455862221)

Fixes: https://github.com/mysociety/caps/issues/480

Screenshot 2023-02-23 at 08 32 02

Screenshot 2023-02-23 at 08 31 34

### DIFF
--- a/scoring/static/scoring/scss/methodology.scss
+++ b/scoring/static/scoring/scss/methodology.scss
@@ -148,7 +148,8 @@ $council-types: (
     'single',
     'district',
     'county',
-    'northern-ireland'
+    'northern-ireland',
+    'combined'
 );
 
 @each $council-type in $council-types {
@@ -214,6 +215,7 @@ $council-types: (
     flex-direction: row;
     flex-wrap: wrap;
     grid-template-columns: repeat(2, 1fr);
+    column-gap: 1rem; // Avoids heavy length content to overlap the next column.
 }
 
 // Tables highlight
--- a/scoring/static/scoring/scss/toggle-section.scss
+++ b/scoring/static/scoring/scss/toggle-section.scss
@@ -71,10 +71,6 @@
     border: 1px solid transparentize($color: $black, $amount: 0.8);
 
     .toggle-section {
-        border: none;
-    }
-
-    .toggle-section + .toggle-section {
         border-top: $border-width solid $border-color;
     }
 }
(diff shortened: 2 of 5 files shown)


lucascumsille:
@zarino I just pushed the latest commits. So now we it's ready for review.

zarino (changes requested):
It seems the council name autocomplete doesn’t work, because the `council_type` value for Combined Authorities in the `window.councils` array is `council_type: "combined"` whereas we’ve used `"combined-authority"` everywhere else (like in the `data-methodology-switch-council-type` attributes and `.js-combined-authority` class).

Otherwise, seems good.

lucascumsille:
@zarino I just pushed a commit that uses `combined` instead of `combined-authority`. Now the search bar and buttons are working. I don't remember if we have a staging site, but should I ask Hannah if I can deploy?

zarino (approved):
Yay, works great now!

Only thing I wonder is whether the table of "section weightings for combined authorities" should have a row per section, the same as the table above it, rather than switching to columns for sections?

Here’s how it’d look:

Screenshot 2023-02-24 at 11 24 30

I’m 50/50. I feel like it makes more sense for the data to be presented in the same way across both tables, but equally, because it’s only a two-column table, it _does_ end up looking a bit weird. Maybe limiting the width of the table would help with that…

Screenshot 2023-02-24 at 11 25 02

Here’s the markup:

```html
<table class="minimal-table" style="max-width: 25em">
    <thead>
        <tr class="bg-green">
            <th>Section</th>
            <th>Combined Authority</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>Transport</td>
            <td>25%</td>
        </tr>
        <tr>
            <td>Buildings, Heating &amp; Green Skills</td>
            <td>25%</td>
        </tr>
        <tr>
            <td>Governance &amp; Finance</td>
            <td>20%</td>
        </tr>
        <tr>
            <td>Planning, Biodiversity and Land Use</td>
            <td>10%</td>
        </tr>
        <tr>
            <td>Collaboration &amp; Engagement</td>
            <td>20%</td>
        </tr>
    </tbody>
</table>
```

lucascumsille:
That was my thought as well =) Using the same pattern that the table above, it seems a bit of a waste of vertical spacing, not massive, but still. Using the full width makes it look weird. So my other two option were: 

- Limiting width, but it breaks the alignment with the table above. 
- So the I tried swapping the columns and the rows. I did think of "Would this swapping be too confusing?", but I thought it's only a 1 row table, don't think it would wreak havoc.

Happy with either option. Let me know which one you'd prefer

lucascumsille:
> My instinct is that the constistency of section names always being in the first column is more valuable than a little bit of saved screen real estate. It makes it clear that the list of sections is different for the two types of council. The two tables having different widths doesn’t feel _that_ bad to me.

Perfect I'll amend that then.

Issue #492, opened 2023-03-07
Update for 2023 Councils


ajparsons:
[Four new councils](https://lite.datasette.io/?url=https://pages.mysociety.org/uk_local_authority_names_and_codes/data/uk_la_future/latest/uk_la_future.sqlite#/uk_la_future/uk_local_authorities_future?_filter_column_1=start-date&_filter_op_1=exact&_filter_value_1=2023-04-01&_filter_column=&_filter_op=exact&_filter_value=&_sort=start-date&_sort_by_desc=on) are being created in April 2023, to replace [20 existing councils](https://lite.datasette.io/?url=https://pages.mysociety.org/uk_local_authority_names_and_codes/data/uk_la_future/latest/uk_la_future.sqlite#/uk_la_future/uk_local_authorities_future?_filter_column=end-date&_filter_op=exact&_filter_value=2023-04-01&_sort=end-date&_sort_by_desc=on).

Underlying data preparation:

- [x] Check CAPE imports are version locked for the following datasets.
- [x] Ensure all required columns are in GovLayers
- [x] Update emissions dataset
- [x] Update deprivation dataset
- [x] Update rurality dataset
- [x] Update distance calculations

Website preparation:

- [x] Import data about councils being replaced into CAPE data structure.
- [x] Add content for former councils referring to current councils.
- [x] Add content for new councils explaining why there's nothing there and linking to old ones.

zarino:
On the new council pages, I’m imagining a fairly prominent box that says something like:

> **This council was formed in [date] by the merging of [number] existing councils**
>
> While we wait for new data about this council, you might want to check out the plans and details of the councils it replaced:
> 
> …

And then lists the replaced councils, with some key information about each, like whether they declared a climate emergency, whether they have any plans/documents, and what their overall 2022 Scorecards score was.

Then on each of the replaced council’s pages:

> **This council was replaced by [new council name] on [date]:**
> 
> …

Followed by the same summary information, but just about the _one_ new (replacement) council.

I _had_ wondered whether an equivalent to the existing [two-tier disambiguation page](https://github.com/mysociety/caps/blob/f3de540a329e746f5a14815c4d5b4c9c0c9a09bc/caps/templates/caps/location_results.html)—but showing both the new and old matching councils for the given postcode—might also help. But maybe, with the changes to the individual council pages, it’s not really required?

ajparsons:
[...]

PR #493, opened 2023-03-08
Prepare for new 2023 Councils

ajparsons:
This PR imports new data (and some new data structures) to cope with the council structure changes in April 2023 (https://github.com/mysociety/caps/issues/492).

- Emissions and similarity datasets have been updated. 
- CAPE given awareness that some councils are no longer 'current'.
- New language to link to predecessor and successor councils.

Notes:
- Needs another pass to sort out those tests.
- I've fixed up the tests for the new 'start-date' column, and to remove the bit that says councils will be removed - but don't quite understand why this has broken the 'number of councils' affected tests. @struan ?

To test:

Run: ```script/update --all``` to improve the new emissions/similarity data. 

To merge:

Sometime after merging in April, the supporting datasets should have the prerelease datasets properly released - and then references here should also be updated. 

## Screenshots

Shows new council, not current county council:

image

Links to former council pages:

image

Links to successor council pages:

image

Postcode goes just to new councils (even when two overlapped before, if both are now gone):

image

image

New emissions data:

image

New similarity data:

image


caps/templates/caps/council_detail.html
```diff
@@ -24,6 +24,39 @@ <h1 class="mt-3 mb-0">{{ council.name }}</h1>
 <div class="row">
     <div class="col-md-7 col-lg-8 order-md-2">
 
+      {% if council.is_old_council %}
```
struan, line 27: I do wonder if we should make these stand out a tiny bit more.

caps/views.py
```diff
@@ -368,7 +373,21 @@ def get_context_data(self, **kwargs):
...
+                if c.replaced_by:
+                    councils.extend(list(c.get_successors()))
+            # remove any councils that are former councils
+            councils = [
+                x
+                for x in councils
+                if x.id in Council.current_councils().values_list("id", flat=True)
+            ]
+            # remove duplicates based on x.id
+            councils = list({x.id: x for x in councils}.values())
```
struan, line 389: if we made councils a set then we could skip this step.

caps/import_utils.py
```diff
@@ -32,7 +93,15 @@ def get_google_sheet_as_csv(key, outfile, sheet_name=None):
...
-def replace_csv_headers(csv_file, new_headers, drop_empty_columns=True, outfile=None):
+def replace_csv_headers(
+    csv_file: PathLike,
+    new_headers: list[str],
+    drop_empty_columns: bool = True,
+    outfile: Optional[PathLike] = None,
+):
+    """
+    Downloading a google sheet csv will sometimes mess up the headers - needs manual fix
+    """
```
struan, line 104: just for clarity could we split the docs/hinting changes into a separate commit?

caps/management/commands/import_emissions_data.py
```diff
@@ -24,7 +24,7 @@ def get_emissions_url() -> str:
     emissions_df = get_dataset_url(
         repo_name="la-emissions-data",
         package_name="uk_local_authority_emissions_data",
-        version_name="1",
+        version_name="1.1.0-futurecouncils",  # replace with '1' at future date
```
struan, line 27: I presume it's expected that this generates a "not the latest version" warning?

struan (changes requested):
A few tiny comments but nothing serious, other than the failing tests.

One thing is that running the emissions update broke combined authority pages because it's finding more than one emission datapoint which isn't the case with the master branch.

caps/models.py
```diff
@@ -209,13 +210,63 @@ class Council(models.Model):
...
+        current_cohort = settings.COUNCILS_AS_OF_DATE
+        return self.end_date is None or self.end_date >= current_cohort
+
+    def is_new_council(self) -> bool:
+        """
+        Is this a new council?
+        """
+        if self.start_date is None:
+            return False
+        return self.start_date >= date(2022, 1, 1)
```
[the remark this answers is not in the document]
ajparsons: Yes - I guess it could be recentish  - so set it to now minus a year?

caps/import_utils.py
```diff
@@ -4,25 +4,86 @@
...
+    ] = np.nan
+    df.loc[
+        get_date_from_from_string(df["end-date"]) >= as_of_date, "end-date"  # type: ignore
+    ] = np.nan
+    df["current-authority"] = df["end-date"].isna() & (
+        df["start-date"].isna()
+        | (get_date_from_from_string(df["start-date"]) < as_of_date)  # type: ignore
+        # set the current-authority correctly
+    )
+    # needs future plannning
```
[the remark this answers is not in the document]
ajparsons: I'll delete that comment - was to remind me to copy the code above into that function so it can handle future data correctly.

caps/management/commands/import_emissions_data.py
ajparsons: Yes, I didn't give it any handling to not warn for prerelease - and tbh it's probably good it does because it'll remind us to change at some point.

ajparsons:
@struan :

> One thing is that running the emissions update broke combined authority pages because it's finding more than one emission datapoint which isn't the case with the master branch.

Interestingly I didn't get this? Will have a closer look.

caps/views.py
ajparsons: I think this case, making a set of django models ends up being unique on memory instances rather than the ID.  Could be wrong about that.

ajparsons:
@struan : should have all data problems fixed. Some parts broken into a different commits, and the tests pass now.

struan (approved):
Looks good. One small thing is that now we have some councils where we don't have the website for and so the website bit of the council page is broken:

Screenshot 2023-04-04 at 11 54 33

This should fix itself shortly but should probably put in some holding text in the meantime.

ajparsons:
Ok, I've added an IF around that. We might need to work together to deploy this one because it'll need to run a fuller version of the update script to get all the new data in.

PR #497, opened 2023-03-17
Reframe “Features” page as “Plans” page, with download and search

### LINKED ISSUE #416: Turn “Features” page to a “Plans” page
There is a high likelihood that we’ll be expanding CAPE out beyond just action plans in the near future. (Technically we already have, with the section on Scottish emissions reduction projects.)

As mentioned in #401, I’m wondering whether the "Browse by feature" page could be repurposed as the new home for everyting to do with plans/documents. Perhaps something like this:

Screenshot 2022-07-14 at 11-47-58 Browse councils by feature

Could also include fairly big plugs for the Scorecards site, and CEUK checklist on this page – given it’s all about finding/making the best plans.

zarino:
- Replace `/features*` URLs with `/plans*`
- Rename “Features” to “Plans” in main nav
- Break homepage plan download and search section into its own partial (plans-download-and-search.html) and include that in both the homepage and the top of the new “Plans” page.
- New `add_context_for_plans_download_and_search` function in views.py, to provide context for the plans-download-and-search.html partial.
- Remove purple hero background for new “Plans” page.
- Add `extra_content_classes` block for adding additional classes to the `site-content` element.

TODO: Handle redirects from /features* => /plans*?

TODO: Pre-warn Vixen about the change, before this goes live? /cc @MyfanwyNixon 

Fixes #416.

Screenshot 2023-03-17 at 10-54-44 Council climate action plans

### DIFF
--- a/caps/static/caps/scss/site-header.scss
+++ b/caps/static/caps/scss/site-header.scss
@@ -10,8 +10,7 @@
         background-color: $color-ceuk-navy;
     }
 
-    .page-purple &,
-    .page-tags & {
+    .page-purple & {
         background-color: $color-ceuk-purple;
     }
 
--- a/caps/static/caps/scss/site-hero.scss
+++ b/caps/static/caps/scss/site-hero.scss
@@ -24,8 +24,7 @@
         background-image: linear-gradient(0deg, darken($color-ceuk-navy, 2%), darken($color-ceuk-navy, 4%));
     }
 
-    .page-purple &,
-    .page-tags & {
+    .page-purple & {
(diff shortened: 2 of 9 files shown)


MyfanwyNixon:
So I think Vixen have pretty much signed off (although they did say they'd still be happy to do odd bits and bobs, my feeling is we can't indefinitely call upon them). I should be able to reflect this change in the ads myself.

MyfanwyNixon:
(although if there are any events being triggered, I _will_ need help from somewhere.

PR #500, opened 2023-04-04
Add semantic search to CAPE

### LINKED ISSUE #494: Prepare CAPE for semantic search text
Get framework in place to accept similar phrases to improve search (from louis's work).

ajparsons:
This PR implements @ldav10's machine learning derived keywords from the CAPE corpus into a semantic search for CAPE. Fixes #494. Also fixes #500.

This ML process created two files:

- A list of key terms present in the overall corpus, and how many documents they appear in.
- A pairwise similarity for each term - that helps draw connections between related terms.

I have manually added a 'highlight' column to the first file for terms we want to draw out and display on council pages.

These are stored in CAPE as a `KeyPhrase` and `KeyPhrasePairWise` models. 

The `import_related_search` management comment loads these files from the [separate github repo](https://github.com/mysociety/cape_ml_search), and then runs a search for each highlighted keyword, and stores which documents matched a specific search. 

Running `script/migrate && script/manage import_related_search` in a local install or codespace should be enough to populate the feature.

# How does the search work?

The 'exact match' tickbox has been replaced by a radio button that includes two new fuzzy matching options. 

image

These do the same thing, but use different thresholds for when a search is a match. Currently these are set to:

```
RELATED_SEARCH_THRESHOLD = 0.6
RELATED_SEARCH_THRESHOLD_LOOSE = 0.5
```

But could be tightened. 

When a search term that is in the overall list of terms is used, a set of related terms above the threshold is retrieved and a new query constructed that is a series of (EXACT_MATCH_ORIGINAL or EXACT_MATCH_ALT1 or EXACT_MATCH_ALT etc).

When these terms are returned, the highlighted text is then slightly cleaned up (merging adjacent highlights), and any of the original related terms that are present in the highlights are returned in addition to the search results. That means any terms that were actually present are highlighted at the top of the search results.

image

# How does the document themes work?

The initial search run on import creates a `CachedSearch` that connects a `KeyPhrase` and `PlanDocument`. These are then used to show beneath a document all terms that have (using the related search approach) matched for this document.

image

Clicking onto a tag opens the search results page limited to that specific document (this is a new option):

image

# Questions

- Are the highlighted terms correct? Are any misleading or duplicated?
- Is the way the new search options are described intuitive enough? What exact descriptions are required?
- Are the thresholds set at the right level to be mostly useful? I'd be tempted to raise to 0.6 and 0.7 just for slightly tighter matches now I've validated it's functionally working. 
- Currently the tags are just thrown underneath a document. There is an un-displayed count that I removed for the moment because it wasn't quite clear what they meant to users at first glance. How should this be displayed?

The number of tags could be changed by:

- Adding a set limit (and sort by number of matches in document?)
- Raising the threshold for relevant terms (meaning less search terms will match per document - but matches will be more meaningful).


caps/forms.py
```diff
@@ -1,14 +1,46 @@
...
+
     council_name = CharField(required=False)
-    exact = BooleanField(required=False)
+
+    MATCH_NORMAL = "normal"
+    MATCH_EXACT = "exact"
+    MATCH_RELATED = "related"
+    MATCH_RELATED_LOOSE = "related_loose"
+
+    MATCH_CHOICES = (
```
zarino, line 33: > Is the way the new search options are described intuitive enough? What exact descriptions are required?

My instinct is, if we think our semantic search is useful, we should make it the default.

But regardless of whether we make it the default or not, we should present the "Search method" options in order of increasing or decreasing specificity, eg:

```
    MATCH_CHOICES = (
        (MATCH_RELATED_LOOSE, "Include as many terms as possible"),
        (MATCH_RELATED, "Include closely related terms (default)"),
        (MATCH_NORMAL, "Search for any of the provided words"),
        (MATCH_EXACT, "Search for the provided phrase exactly"),
    )
```

And the second option would be selected by default.

caps/templates/caps/search_results.html
```diff
@@ -55,6 +66,16 @@ <h1 class="mb-3">Search inside plans</h1>
     <div class="col-md-8">
       {% if query %}
 
+      <div class="card ceuk-card ceuk-card--red ceuk-card--muted mb-3 mb-lg-4">
+        <div class="card-body mb-n3">
+          <h2>Search results for "{{ query }}"</h2>
+          {% if related_words %}
+            <p>Terms like the following were also used to improve the search: {% for term in related_words %}<span class="badge badge-secondary">{{ term }}</span> {% endfor %}</p>
```
zarino, line 73: In your comment, you say:

> That means any terms that were actually present are highlighted at the top of the search results.

If that’s the case, why does this template say "Terms _like_ the following were also used…"? Isn’t it just "The following terms were also used…"?

zarino: Also, no need for these to be marked up as `<span class="badge">`. Looks super heavy when there are dozens of them in a massive block. And people might expect to click them, because they look like little buttons.

caps/templates/caps/council_cards/council-documents.html
```diff
@@ -0,0 +1,130 @@
...
+                                                    {% endif %}
+                                                </p>
+                                            </div>
+                                            <div></div>
+                                            {% if plandocument.is_deleted %}
+                                                {% if plandocument.file_exists %}</a>{% endif %}
+                                            {% else %}
+                                            </a>
+                                        </div>
+                                        {% if plandocument.sorted_key_terms %}
```
zarino, line 102: Is there a reason you think we should include this list of terms below each document?

From a quick test here, it looks like most documents have at least a few dozen terms in them, which makes this "wall" of tags pretty difficult to parse or make use of.

I think I’d either:

1. Show only the most frequent terms in each document (along with the number of times they appear) – although, TBH, I’m still not sure how useful this will be in reality.
2. Remove this listing of terms entirely, but pull out the most frequent terms across _all documents_ for the council, and present them as search suggestions under the search box:

Screenshot 2023-04-04 at 11 23 15

In either case, we have to set a cut-off somewhere for how many tags we want to display, because I really don’t think more than a handful will make sense.

The other alternative, of course, is that we just remove these bits _entirely_ from the council detail page, and use them _only_ for semantic search on the search page.

caps/templates/caps/search_results.html
```diff
@@ -36,11 +31,27 @@ <h1 class="mb-3">Search inside plans</h1>
...
                         <label for="council_name" class="d-flex flex-wrap flex-row-reverse">
                             <span class="text-muted">(Optional)</span>
-                            <span class="mr-auto">Council name contains</span>
+                            <span class="mr-auto">Council name</span>
                         </label>
                         <input type="text" class="form-control" name="council_name" id="council_name" value="{{ form.council_name.value|default_if_none:'' }}">
                     </div>
                   {% endif %}
+
+                    <p>Search method</p>
```
zarino, line 40: FYI radio buttons should be grouped inside a `<fieldset>`, eg:

```html
<fieldset class="form-group">
    <legend style="font-size: 1em; font-weight: bold;">Search method</legend>
    <div class="form-check">
        <input … >
        <label … >
    </div>
    …
</fieldset>
```

Sorry it looks like we have pretty obnoxious default styling on `<legend>` elements, hence the inline styles in the code above. We should probably just fix the default. (Upgrading the site to Bootstrap 5 would help with this, and is also been something I’ve been meaning to do for ages.)

zarino (commented):
I’ve left a few comments about presentation – although I’d appreciate your input on whether they make sense!

caps/templates/caps/council_cards/council-documents.html
ajparsons: So I think this is actually the really cool feature I think - possibly more useful that the search improvement itself.

It's not so much the little tags, but the fact you can click through and the search is letting you see the relevant bit inside that document. This is why a reasonable amount of tags isn't necessarily bad (but probably fewer than they are), because it's all jumping off points to inside the documents. 

Basically, this is making it easier for visitors to more quickly jump into bits of a plan, rather than hitting a 'now download this pdf, then look inside the pdf' wall.

caps/templates/caps/search_results.html
ajparsons: Yes, that was from an earlier bit where I was showing terms that weren't necessarily matching. Should be changed.

struan (changes requested):
I've not looked at all of this in detail but I'm not sure that `get_sematic_query` is doing quite the right thing. It all does seem to do what I expect though.

caps/forms.py
```diff
@@ -56,7 +110,19 @@ def search(self):
...
+        result = sqs.highlight(**highlighter_kwargs)
+
+        if self.cleaned_data["document_id"]:
+            # adjust the returned object to only include matches for the specific document
+            result = [x for x in result if x.pk == self.cleaned_data["document_id"]]
+
+        if self.cleaned_data["match_method"] != HighlightedSearchForm.MATCH_NORMAL:
+            # improve the highlighting by removing the <mark> tags from the middle of words
+            # for all searches that use exact text
+            result, related_terms = condense_highlights(result, related_terms)
```
struan, line 126: This is interacting poorly with the pagination which means it does a query for each results page every rather than a single query. I'm not entirely sure how to fix it at the moment.

caps/models.py
```diff
@@ -1262,3 +1265,112 @@ class Meta:
...
+        """
+        return cls.objects.filter(highlight=True)
+
+    def display_name(self) -> str:
+        """
+        Get the name to display for this keyphrase
+        """
+        return self.nice_phrase or self.keyphrase
+
+    def related_phrases(
```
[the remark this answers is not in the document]
ajparsons: No, it all ended up being done in bigger queries in the end. Could be deleted.

caps/search_funcs.py
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    query_parts = []
+    all_related_terms = []
+
+    def get_word_with_cosine(x: KeyPhrasePairWise) -> Optional[KeyPhrase]:
+        """
+        Get back to the phrase object we already have, but keep the cosine
+        from the comparison
+        """
```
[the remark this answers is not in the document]
ajparsons: This is solving a problem where we want the Keyphrase object, but the similarity is stored on the pairwise comparison.

Later on, we want to be able to sort by the similarity when showing which related terms were used - so we can highlight the less expected terms that have been used (e.g. conceptually different, rather than 'light typo').

caps/search_funcs.py
```diff
@@ -0,0 +1,171 @@
...
+    Run a semantic search query against the contents of the plan documents
+    """
+
+    # get all the related terms
+    all_related = list(KeyPhrase.objects.all())
+    all_related = [x for x in all_related if len(x.keyphrase) > 4]
+    all_related_lookup = {x.id: x for x in all_related}
+    triggered_keywords = []
+    for related in all_related:
+        if related.keyphrase.lower() in q.lower():
```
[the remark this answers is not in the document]
ajparsons: Yep.

caps/search_funcs.py
```diff
@@ -0,0 +1,171 @@
...
+    q: str,
+    threshold: float = settings.RELATED_SEARCH_THRESHOLD,
+) -> tuple[Optional[SearchQuerySet], list[KeyPhraseSearch]]:
+    """
+    Run a semantic search query against the contents of the plan documents
+    """
+
+    # get all the related terms
+    all_related = list(KeyPhrase.objects.all())
+    all_related = [x for x in all_related if len(x.keyphrase) > 4]
```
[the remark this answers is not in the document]
ajparsons: I was adjusting after I'd done a big import - but yes, I can apply this to the import as well.

caps/search_funcs.py
```diff
@@ -0,0 +1,171 @@
...
+        query_parts = Q(text__iexact=Exact(q))
+        for term in related_terms:
+            alt_text = q.replace(keyword.keyphrase, term.keyphrase)
+            query_parts |= Q(text__iexact=Exact(alt_text))
+        # run the query
+        sqs = SearchQuerySet().filter(query_parts)
+
+        all_related_terms = list(set(all_related_terms))
+
+        return sqs, all_related_terms
```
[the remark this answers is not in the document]
ajparsons: Yep, that's an error. Will fix up. It hasn't been obvious because I've just been looking at one term.

caps/forms.py
ajparsons: I think I've fixed this by moving the highlighting adjudgment after pagination has happened.

caps/search_funcs.py
```diff
@@ -0,0 +1,171 @@
...
+    threshold: float = settings.RELATED_SEARCH_THRESHOLD,
+) -> tuple[Optional[SearchQuerySet], list[KeyPhraseSearch]]:
+    """
+    Run a semantic search query against the contents of the plan documents
+    """
+
+    # get all the related terms
+    all_related = list(KeyPhrase.objects.all())
+    all_related = [x for x in all_related if len(x.keyphrase) > 4]
+    all_related_lookup = {x.id: x for x in all_related}
```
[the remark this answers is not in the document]
ajparsons: I've moved the lookup creation

caps/search_funcs.py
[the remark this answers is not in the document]
ajparsons: Done this.

caps/management/commands/import_related_searches.py
```diff
@@ -0,0 +1,91 @@
...
+                count = highlighted.count("<mark>")
+            else:
+                count = 0
+            tqdm.write(f"{doc_id} {count} {keyphrase.keyphrase}")
+            if count > 0:
+                new_searches.append(
+                    CachedSearch(search_term=keyphrase, document_id=doc_id, count=count)
+                )
+
+    CachedSearch.objects.all().delete()
```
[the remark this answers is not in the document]
ajparsons: I was struggling to find a way to query inside a specific document (despite it obviously knowing about individual documents) - but yes, if you figure that out, this could be limited to a subset fairly easily.

caps/templates/caps/search_results.html
ajparsons: Made this change.

caps/templates/caps/search_results.html
ajparsons: now looks like this 
image

caps/search_funcs.py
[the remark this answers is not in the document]
ajparsons: Should be resolved.

zarino:
@ajparsons I think I’m happy with this from a UI point of view. But would be open to changes, if you check it out and feel that something’s amiss.

ajparsons:
Lovely 
image

Just going to have a pass and see if I can address one of Struan's points I put off last time.

ajparsons:
@struan : I've made the default behaviour without '--all' just doing new docs from the last few weeks rather than all documents.
The trick is 'django_id' is what's tracking the internal object ids. 

```
    # if given a specific set of documents limit to that
    if limit_to_ids:
        sqs = sqs.filter_and(django_id__in=limit_to_ids)
```

caps/management/commands/import_related_searches.py
```diff
@@ -0,0 +1,116 @@
...
+            "--reload-searches",
+            action="store_true",
+            help="Runs the search against all documents",
+        )
+
+    def handle(self, *args, **options):
+        get_all = options["all"]
+        reload_searches = options["reload_searches"]
+        if get_all or not KeyPhrase.objects.count():
+            print("Downloading data")
```
struan, line 106: It would be nice if these were only printed with a verbose option

struan (approved):
One small request for the import output but otherwise good.

caps/management/commands/import_related_searches.py
ajparsons: Added a quiet option that disables progress bars, etc - and made that the default for script/update

PR #504, opened 2023-04-21
Report events using GA4 Measurement Protocol

ajparsons:
Problem described in https://github.com/mysociety/local-intelligence-hub/issues/243 - but:

- We want to use GA4, but not use cookies.
- If you do this, GA4 doesn't send custom events for some reason.
- You *can* send events separately through [the measurement protocol](https://developers.google.com/analytics/devguides/collection/protocol/ga4/reference?client_type=gtag). 

What complicates this is that the measurement protocol uses a separate secret key - which keeping secret needs some server-side help. It's unclear to me what the consequences of losing the key would be (people could send events to you - but I think you could do that anyway through gtag I think?).

My approach:

- Add the measurement api key to env variables.
- Write a django function that takes a POST request and sends the event off to google.
- Write javascript function that sends event from browser to our server, to be sent on to google.

The browser doesn't know the secret, and the csfr token protects the endpoint a bit (but again, unclear what the worst case is here).

If you decided you didn't care about the secret being public, you could condense all this into the javascript function. 

Looks like this:

```mermaid
sequenceDiagram
    Google->>Django: api_secret (env var)
    Django->>Browser: csrf_token
    Google->>Browser: client_id
    Browser->>Django: csrf_token, client_id, event
    Django->>Google: api_secret, client_id, event
```


ajparsons:
Ah, so google's line:

> We recommend that you keep these private to your organization. If you deploy the measurement protocol client-side, you should regularly rotate api_secrets to avoid excessive SPAM.

So we could do that if we wanted and get it all into the javascript.

zarino:
Aside from writing events, can the Measurement API key be used for anything else? Like _reading_ events? If so, we’ll want to think twice about making it public.

But if not, it definitely feels like it’s worth simplifying to a client-only implementation, if we can. Would make it _so_ much easier to implement in our other sites, as a drop-in JavaScript library.

ajparsons:
I can't see any reason why we *currently* wouldn't mind that.

I guess the risk is that at a later point they decide that because it's secret, it can be used for more things.

ajparsons:
Closing in favour of #506

PR #508, opened 2023-05-03
Fix new council issues

### LINKED ISSUE #507: New (2023) councils showing as zero-scores at the top (!!) of the 2021 Scorecards tables
Title says it all really. New (2023) councils are showing weirdly on the 2021 Scorecards site:

Screenshot 2023-05-03 at 09 40 52

My instinct is the Scorecards site should be completely ignoring these new councils – they didn’t exist at the time the 2021 Scorecards were collated, and even if we allowed people to search for them (because people might want to look up their _current_ council) there’s not much we can show them, as their council won’t have been scored.

However, if that causes complications, then we should sort these councils to the bottom of the table and replace their empty scores with a message, like we do for the councils that didn’t have a plan at the time of marking. (In this case, the message would be "This council did not exist when all plans were assessed on 20th September 2021.")

struan:
Deals with scorecards table sort order, council page ans council comparisons for post April 2023 councils.

Fixes #507

### DIFF
--- a/scoring/templates/scoring/council.html
+++ b/scoring/templates/scoring/council.html
@@ -26,6 +26,9 @@
         <div class="col-12 col-lg-5">
             <div class="about-section-child pt-3">
                 <h3 class="mb-3">About this council</h3>
+                {% if new_council %}
+                <p>This council was created on April 1, 2023 so was not included in the 2021 Scorecards.</p>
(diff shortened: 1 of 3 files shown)


zarino (approved):
Homepage and council pages both look great 👍

PR #531, opened 2023-06-21
[WIP] Council climate twins page

### LINKED ISSUE #530: Create experimental “find your council twin” interface
- Search input and autocomplete (same as homepage)
- Results table
- Banner ad on council pages, bypassing the search step

zarino:
Fixes #530.

TODO:

- Working semantic topic extraction
- Working public opinion data
- Detailed data in the “extended family” table

Screenshot 2023-06-23 at 16-18-46 Liverpool City Council’s climate twin


ajparsons:
Okay! So I've got some commits in there that make the keyphrase comparison functional.

My bigger thoughts:

* I think you want to go wider that just 'one' twin. This is quite easy to adapt wider, and you could have the related text at the bottom link to alternative views. 
* The big reason for this is the nearest 'twin' might not be surprising, and so might not be that informative. But being able to see more info on differences across a range (as you do at the bottom) might be!
* Along with that, 'twin' language might need to change.
* For technical reasons, the polling isn't good for a comparison. Core reason is these technically have error margins, but I don't know them, and so small differencesmay not be statistically significant. Because MRP polling uses demographics to create local results, and similarity is calculated in ways that might reflect similar demographics, it is likely that most 'twin' councils will have similar polling results, and in turn that that difference is meaningless.

ajparsons:
Also the possibly unbound twin variable was bugging me because I couldn't think of a better way to do it, it turns out you can use next to grab the first one, or None if there's nothing:

```
twin = next(
    (
        group["councils"][0]
        for group in related_councils
        if group["type"].slug == "composite"
    ),
    None,
)

if twin in None:
    raise ValueError("No valid twin, no similar composite councils calculated")
```

Won't change myself because it's not necessarily more readable.

PR #535, opened 2023-07-03
fix missing document title for plans with no title

### LINKED ISSUE #534: Missing document titles on CAPE council pages
When #500 was merged, it brought with it a bug that wasn’t obvious in local development:

Screenshot 2023-06-30 at 16 53 17

(eg: https://cape.mysociety.org/councils/welwyn-hatfield-council/)

I’ve noticed this on a number of council pages, but not all. So there must be something about those specific documents which means the title isn’t working. Did we accidentally remove something that was autogenerating titles, when we refactored the climate documents display code?

struan:
Fixes #534

### DIFF
--- a/caps/templates/caps/includes/council-document.html
+++ b/caps/templates/caps/includes/council-document.html
@@ -22,17 +22,17 @@
 
     <div class="council-document__info">
 
+        {% if plandocument.title %}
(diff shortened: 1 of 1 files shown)


zarino (approved):
Woop! Thanks for spotting!

Issue #515, opened 2023-05-15
Council pages for 2023 Scorecards


zarino:
Almost the same content and layout as last year’s council page, with a few tweaks. Full page description from top to bottom:

### Hero

Council name and type, in the hero, as on the 2022 council pages.

### “About this council” section

If this council was _also_ marked in the 2022 Plan Scorecards, their 2022 Plan Scorecards score and a link to their 2022 Plan Scorecards page should be added to the "About this council" section.

### “About the scoring” section

Wording changes TBC by CEUK.

This may also include a "Key definitions" list, with definitions of terms like "score" and "weighting", with a link to the Methodology for more info.

### “Scorecard section performance” section

Same as last year – a table with a row per section, and a column for the section name, this current council’s score, and the average score across councils of the current type.

**Note** Combined Authorities have different sections to the other council types, so we need to handle that.

### “Individual question performance” section

Add controls to allow filtering the table to show only questions that:

- Were marked from a particular source (FOI / volunteer / national data / volunteer + national data)
- Are "council operations only"

Then a table of questions for this council, with a row per question (grouped into sections, like the current fold-out question table on the single council page) and columns for question number, question text (with link to show criteria), question source (see above), question weighting, this council’s score on each question/section, score for a chosen comparison council, number of councils of the same type who got full marks.

**Note** that this means there still needs to be a way to add up to 1 additional council to compare, as a new column, in this table (as on the 2022 council pages).

**Note** that a small number of the questions will be scored with negative marks!! For these questions, the marks will be displayed as either:

- “No negative points awarded” (instead of “0” which could be misunderstood)
- “-1” / “-2” / “-3” (if negative points awarded)

**Note** that the table won’t have `.subpoint-question` rows any more – just collapsible sections, with a single level of questions inside.

lucascumsille:
Almost the same content and layout as last year’s council page, with a few tweaks. Full page description from top to bottom:

### Hero

- [ ] Council name and type, in the hero, as on the 2022 council pages.

### “About this council” section

- [ ] If this council was _also_ marked in the 2022 Plan Scorecards, their 2022 Plan Scorecards score and a link to their 2022 Plan Scorecards page should be added to the "About this council" section.

### “About the scoring” section

Wording changes TBC by CEUK.

- [ ] This may also include a "Key definitions" list, with definitions of terms like "score" and "weighting", with a link to the Methodology for more info.

### “Scorecard section performance” section

[...]

PR #544, opened 2023-08-31
2023 Scorecards council page

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/515

@zarino not sure if we should merge on master or the 2023-initial-views. 

I tried to got rid of much unnecessary code as possible.

Almost the same content and layout as last year’s council page, with a few tweaks. Full page description from top to bottom:

### Hero

- [x] Council name and type, in the hero, as on the 2022 council pages.

### “About this council” section

- [x] If this council was _also_ marked in the 2022 Plan Scorecards, their 2022 Plan Scorecards score and a link to their 2022 Plan Scorecards page should be added to the "About this council" section.

### “About the scoring” section

Wording changes TBC by CEUK.

- [x] This may also include a "Key definitions" list, with definitions of terms like "score" and "weighting", with a link to the Methodology for more info.

### “Scorecard section performance” section

- [x] Same as last year – a table with a row per section, and a column for the section name, this current council’s score, and the average score across councils of the current type.
- [ ] The Autocomplete is working now, but with the reset filter I get this error:

Image

**Note** Combined Authorities have different sections to the other council types, so we need to handle that.

### “Individual question performance” section

- [x] Add controls to allow filtering the table to show only questions that:
  - [x] Were marked from a particular source (FOI / volunteer / national data / volunteer + national data)
  - [x]  Are "council operations only"

- [x] Then a table of questions for this council, with a row per question (grouped into sections, like the current fold-out question table on the single council page) and columns for question number, question text (with link to show criteria), question source (see above), question weighting, this council’s score on each question/section, score for a chosen comparison council, number of councils of the same type who got full marks.

- [x] **Note** that this means there still needs to be a way to add up to 1 additional council to compare, as a new column, in this table (as on the 2022 council pages).

- [x] **Note** that a small number of the questions will be scored with negative marks!! For these questions, the marks will be displayed as either:

- [x] “No negative points awarded” (instead of “0” which could be misunderstood)
- [x] “-1” / “-2” / “-3” (if negative points awarded)

- [x] **Note** that the table won’t have `.subpoint-question` rows any more – just collapsible sections, with a single level of questions inside.

### Important from a UX point of view there are a couple of things that would be nice to work:

- [ ] I left some TODO for Struan, things like if the question is council operational only do this. Or add classes to the row.


lucascumsille:
@zarino I'll be adding the stuff we talked about in the CEUK/Design meeting to this PR

lucascumsille:
@zarino In[ this commit](https://github.com/mysociety/caps/pull/544/commits/ef4c92c6f38523cda33fac841496917983a5ea95) I have included a list with the changes made. Let me know if I forgot something. =)

zarino (approved):
@lucascumsille I hope you don’t mind, but I’ve added three commits proposing some changes here.

The section and question rows felt very messy to me – the vertical stacking of the Source / Weight / Evidence bits, the box-within-a-box red alerts for negative questions / penalties, the "badges" around so many elements…

Screenshot 2023-09-15 at 12 46 28


So I’ve added a commit in which I’ve tried to simplify it all down…

Screenshot 2023-09-15 at 12 47 26


As part of this, the Source, Weight, and Evidence are all hidden within a `<details>` element – we’ll need to see whether CEUK are happy with this. (Considering they _originally_ wanted an always-visible column for each of these!)…

Screenshot 2023-09-15 at 12 49 01


Then I added a commit which pops the table filter/comparison controls out of their accordion and into two always-visible cards – again, trying to simplify things a bit. My worry is, if we don’t make these things visible at first pageload, people won’t discover them. For the cards, I’ve used the same markup as your proposed "feature" cards in the [`scorecards-2023-homepage-features`](https://github.com/mysociety/caps/compare/scorecards-2023...scorecards-2023-homepage-features)…

Screenshot 2023-09-15 at 16 28 19


Still, with all these changes made, I was annoyed that the question table itself was so far down the page. I saw you’d left in two options for the layout at the top of this page, but I’ve added a commit that does something perhaps a bit more contentious – moving the council statistics into the header, and then places the About / Key terms stuff side-by-side, again removing yet another accordion:

Screenshot 2023-09-15 at 16 28 48


My thinking behind this change was to add some clear zoning to the page – the hero is all the background info about the council, the next section is all the background info about the _Scorecards project_ as a whole, and then the final section is this council’s actual scores.

While I was making these changes, I also re-indented the markup ([we try to](https://pages.mysociety.org/coding-standards.html) use a 4-space indent in Django templates, but with Django blocks like `{% if %}` indented 2 spaces away from the thing they wrap around – it tends to make the nested Django and HTML structure easier to spot).

I’m aware you have some outstanding fixup items listed on this PR, but I figure—assuming you’re happy with my suggested changes—we merge this branch as-is, and move those outstanding fixes into [your Nice-to-haves ticket](https://github.com/mysociety/caps/issues/545).

lucascumsille:
Thanks for the feedback @zarino 
A few things:

- Not a superfan of the hero section, seems messy, not really easy to scan. Probably it's the length difference between short words like "Nation" and "Index of multiple deprivation"
- The key definition section was an accordion, because it's unclear it's total length of the keywords section, so there could be a scenario when the right section is way taller than the left section, but happy to come back to that once we have the copy, alternative if it's too long we could have a 'show more' . At the same time that section might morph into a headline/highlights section, which could justify removing the accordion. but we might still wanting to move it to a more noticeable area of the page, considering is a highlight/headline section.
-Agree, I would check with CEUK if they are fine having the 'how is marked?' hidden by default.

lucascumsille:
@zarino I'll update this page with what CEUK has just mention:

- [ ] Negative section with red just for the alert and not the row.
- [ ] Modal for criteria
- [ ] Add a checkbox to unfold every summary.

zarino:
> Negative section with red just for the alert and not the row.

FYI @lucascumsille, Isaac’s request here was that the section row only be fully red if the council’s _total score for that section_ is negative (less than 0) – otherwise, if they received a penalty in that section but it didn’t bring their score below 0, then the section row stays pale purple, but the "negative score" warning message _inside_ the row stays red.

Contrast between the red and purple, in that second case, is going to be a challenge (this is the reason I removed the two colours in the first place ;-) ) and adding a border might make the warning look like a button. I guess, try things until it looks ok!

lucascumsille:
Thanks @zarino 
I just added a few fixups :

### Criteria
I tried to use the loop for the IDs that we use in the methodology 2023 page, but It didn't work. Any ideas how I can fix that? not sure if it's because I also need to change something in `views.py`

Screenshot 2023-09-19 at 07 50 14

### Negative section
When incur in a penalty:
Screenshot 2023-09-19 at 06 34 04

When negative points:
Screenshot 2023-09-19 at 06 32 16

Let me know what you think, otherwise we could just leave the section with white background by default and only adding a red background when there is negative point.

### Checkbox to unfold 'How it was marked?'
I'll use the same ID loop from the criteria section.

lucascumsille:
@zarino Here is a preview of how it looks with the latest changes:


Screenshot 2023-09-27 at 12 19 42
Screenshot 2023-09-27 at 12 19 38
Screenshot 2023-09-19 at 07 50 14
Screenshot 2023-09-27 at 12 19 58

zarino (approved):
Looks good @lucascumsille :+1:

lucascumsille:
@zarino and @struan I can see the comparison tool is working, but when I pull the latest branch this morning some of the styling was broken, so I have added some extra commits to fix the "Criteria" and "How is this marked" section + some minor styling stuff.

One thing is pending, but not sure if you prefer to tackle it after merging: It is to add the criteria. Currently is a placeholder. I tried to replicate it what we did in the methodology page, but instead of `{{ question.criteria}}`, I used  `{{ answer.criteria}}`, but I don't think that's enough. (I didn't commit this)

Let me know if you are happy to merge this.

PR #548, opened 2023-09-07
Add 'Your Representatives' box to council page

ajparsons:
This PR adds 'Your Representatives' box to council page, which links to WriteToThem.

When doing a postcode search, a session cookie will be generated with the postcode, which is then used to pre-populate the form. Form also works fine without this.

With cookie:

image

Without cookie:

image

Also some small tweaks elsewhere:

- Reduced frequency of initial survey (25% chance rather than 100% on second page view).
- Light change to description of local polling.

### DIFF
--- a/caps/helpers/council_navigation.py
+++ b/caps/helpers/council_navigation.py
@@ -44,6 +44,12 @@ council_menu = [
         desc="What the council is responsible for and how it relates to climate change.",
     ),
     MenuItem(
+        slug="representatives",
+        title="Your representatives",
+        color="blue",
+        desc="Find your representatives.",
+    ),
+    MenuItem(
         slug="declarations",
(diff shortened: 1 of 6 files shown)


caps/templates/caps/base.html
```diff
@@ -83,7 +83,7 @@
     </div>
 
     {% include 'caps/includes/site-footer.html' %}
-
+
```
struan, line 86: can we remove this purely whitespace change from the commit?

struan (approved):
Looks good.

One tiny for the sake of neatness comment.

caps/templates/caps/base.html
ajparsons: Yep!

PR #553, opened 2023-09-18
Scorecards 2023 home page data

struan:
Display 2023 data on the Scorecards front page.

This includes the data import for the overall and section scores.


struan:
This has been merged

PR #556, opened 2023-09-19
Update plans import approach

ajparsons:
I've changed the google sheet to a new simpler format, given it now just powers CAPE. 

IMPORTANT: the import will not work until this has merged because I've done this. 

This also will interpret a 'True' in the `out_of_date` column to remove from the import, letting us keep track of changes in the sheet. 

Few other loose changes:

- Added retry approach to scraper, seems to get a few of the docs it was struggling with before.
- Fixed link to document from search page - which didn't work for non-pdfs.

### DIFF
--- a/caps/management/commands/preprocess.py
+++ b/caps/management/commands/preprocess.py
@@ -1,6 +1,7 @@
 import os
 import ssl
 import sys
+from functools import lru_cache
 from os.path import basename, isfile, join, splitext
 from urllib.parse import urlparse
 
@@ -10,6 +11,9 @@ import requests
 import urllib3
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+
 from caps.import_utils import get_google_sheet_as_csv, replace_csv_headers
 from caps.models import PlanDocument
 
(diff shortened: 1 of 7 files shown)


caps/management/commands/preprocess.py
```diff
@@ -145,26 +145,51 @@ def replace_headers():
...
-            "scope",
-            "status",
-            "homepage_mention",
-            "dedicated_page",
-            "well_presented",
-            "baseline_analysis",
-            "notes",
-            "plan_due",
             "title",
-            "title_checked",
```
struan, line 146: This row was used to see if the title should be updated. without it all the document title's get set to blank.

The `get_plan_defaults_from_row` function in `import_plans` will need to be updated to account for this.

struan (changes requested):
One bug, but otherwise this seems to work.

There's much less information being imported but I guess if this new sheet can be updated automatically it's overall better. Plus we weren't really using the missing info.

caps/management/commands/preprocess.py
struan: The test is `if char_from_text(row["title_checked"]).lower() == "y":` so Y or y would work

ajparsons: Have removed what was checking this later on.

PR #560, opened 2023-09-27
Scorecards 2023 homepage features layout

lucascumsille:
Screenshot 2023-09-27 at 12 27 34

New horizontal layout for homepage features.

### DIFF
--- a/scoring/templates/scoring/home.html
+++ b/scoring/templates/scoring/home.html
@@ -62,54 +62,42 @@
         </div> {% endcomment %}
 
         <div class="row">
-            <div class="col-12 col-lg-5">
-                <div class="accordion" id="accordionMainMenu">
-                    <div class="accordion-item">
-                      <h2 class="accordion-header" id="mainMenuHeadingOne">
-                        <button class="accordion-button" type="button" data-bs-toggle="collapse" data-bs-target="#mainMenuCollapseOne" aria-expanded="true" aria-controls="mainMenuCollapseOne">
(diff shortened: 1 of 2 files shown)


lucascumsille:
@zarino Sorry I thought I had made this PR already.

zarino (approved):
Layout looks good to me. There’s some wording tweaks I think we should make ("Combined Authorities" instead of "Combined" in the `main-filter.html` partial, the use of "Review" in two of the headings, seems an odd choice to me) but all of the text here is provisional anyway, so I figure we can change it all in one go.

lucascumsille:
@zarino I just updated the copy:

- Combines -> Combined Authority
- Review -> Check (However I'm not too bothered by this, because most likely CEUK will add their own copy anyways, now that I gave the screenshot to Annie.)

lucascumsille:
Let me know if you want me to merge, once I had done the rebase.

zarino (approved):
Good enough to merge for now, rather than holding this up.

But looking at it in place, I think we need to make it clearer that "Filter Scorecard table by type of council" is _connected_ to the table below – like, the fact that the "Single Tier" button is selected (purple) is _the reason_ that you’re seeing what table you’re seeing, and when you type into the "Council name" search box above the row of buttons, it’ll do something with the table below. 

Also, don’t think we need the "Scorecards results" heading above the three boxes either.

But as I say, let’s just get this merged, and we can discuss and revisit with CEUK later.

Issue #514, opened 2023-05-15
Move 2022 Plan Scorecards to subdirectory


zarino:
The 2022 Plan Scorecards pages should all move to:

- `/plan-scorecards-2022/`
- `/plan-scorecards-2022/scoring/<council-type>/`
- `/plan-scorecards-2022/council/<council-name>/`
- `/plan-scorecards-2022/questions/<question-slug>/`
- `/plan-scorecards-2022/methodology/` (note, the 2022 methodology, _not_ the 2023 draft!)
- `/plan-scorecards-2022/how-to-use-the-scorecards/`
- `/plan-scorecards-2022/about/`
- `/plan-scorecards-2022/contact/`

Also, a few tweaks to be made to the top of the 2022 homepage and “scoring” table pages:

- Remove Yellow “Now available: Council Climate Action Scorecards draft methodology” banner.
- Add something that makes it _really clear_ you’re looking at the outdated / frozen-in-time 2022 site, and that the more up-to-date 2023 site is a click away.

lucascumsille:
> Add something that makes it really clear you’re looking at the outdated / frozen-in-time 2022 site, and that the more up-to-date 2023 site is a click away.

Initial ideas:

- Subnav banner advising the user.
- Change the BG navcolour between old VS new scorecards.

lucascumsille:
Note to self: Add the switch to the 2021 version

PR #563, opened 2023-09-28
Scorecards 2021 navbar and Hero section

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/514

This PR includes:

- Scorecard version selection to the navigation bar 2021
- New hero section styling to match 2023, so users can easily identify which version of the scorecards they are seeing.

### Preview
Screenshot 2023-09-28 at 16 10 21
Screenshot 2023-09-28 at 16 10 15

### DIFF
--- a/scoring/templates/scoring/base.html
+++ b/scoring/templates/scoring/base.html
@@ -50,10 +50,8 @@
                         2023
                     </a>
                     <ul class="dropdown-menu position-absolute ms-n3 shadow-lg" aria-labelledby="scorecardsVersion">
-                        <!-- TODO: Link to current page -->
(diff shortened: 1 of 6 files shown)


zarino (changes requested):
A couple of comments inline.

Have tried out the dropdown UI and it seems fine on desktop and mobile 👍 

One final thing – while you’re editing `scoring2022/home.html`, I think you could get rid of the yellow section at the top entirely – it won’t need to be there once the new scorecards have launched.

lucascumsille:
Thanks @zarino I just added to fixups to tackle all the above. Also included those changes for the 2023 version.

zarino (approved):
:+1:

Issue #503, opened 2023-04-18
Write privacy policies for Scorecards and GRACE


zarino:
Separate PPs for Scorecards and GRACE, linked to from the site footers somewhere.

Don’t need to be anything ground-breaking. Use [Local Intelligence Hub privacy policy](https://github.com/mysociety/local-intelligence-hub/blob/94a8c252b523d510d16f92bf2bb8f24b160a96f3/hub/templates/hub/privacy.html) as an example. There is more info about privacy policy requirements on the mySociety wiki.

zarino:
Draft Scorecards privacy policy:

https://docs.google.com/document/d/1vSe9-cSRbrMvsLEDn4oxKJOF4_-1CXfsMs12iPxAPkg/edit

Draft GRACE privacy policy:

https://docs.google.com/document/d/1vmw5Va5nwoA2u2ahefSg6zLD246DL280Q2yaAFz6ZnQ/edit

Need review from CEUK and Louise/Sam.

zarino:
Discussed this today with @sagepe. Summary:

- @sagepe to go away and think about whether mySociety should be described as a joint controller with CEUK, or whether CEUK are the controller and mySociety are the processor.
   - Note for future – we may want to include data controller/processor decision in our future MOUs with partners, to make this clearer from the outset.
- @zarino to check missing a data retention period for GRACE user accounts data. Could we delete the accounts data (but probably keep the accounts themselves, just without emails or names) once the Scorecards are published?

Aside from those two issues, both drafts seem fine.

@sagepe also noted that we’re missing a privacy policy for CAPE. I’ve created a ticket for that here: https://github.com/mysociety/caps/issues/554

zarino:
@sagepe I‘ve filled in the top half of the [GRACE legitimate interests test wiki page](https://wiki.mysociety.org/wiki/GRACE_GDPR_Legitimate_Interests_test) – let me know if there are any changes you think I should make!

sagepe:
I've put a comment on [the Talk page](https://wiki.mysociety.org/wiki/Talk:GRACE_GDPR_Legitimate_Interests_test), as I think it would be good to track that discussion there for the sake of a complete record.

zarino:
The Scorecards privacy policy has been [merged into the `scorecards-2023` branch](https://github.com/mysociety/caps/pull/569), so will go live with that on 18th October.

[...]

PR #569, opened 2023-10-04
Scorecards 2023 - Privacy policy page

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/503

Preview here:

https://github.com/mysociety/caps/assets/13790153/c83873bf-6d12-4bf1-a9f5-a88f3ff20746

### DIFF
--- a/scoring/views.py
+++ b/scoring/views.py
@@ -47,6 +47,15 @@ class LogoutView(LogoutView):
     next_page = "scoring:home"
 
 
+class PrivacyView(TemplateView):
+    template_name = "scoring/privacy.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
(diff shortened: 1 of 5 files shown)


scoring/views.py
```diff
@@ -47,6 +47,10 @@ class LogoutView(LogoutView):
     next_page = "scoring:home"
 
 
+class PrivacyView(TemplateView):
+    template_name = "scoring/privacy.html"
```
zarino, line 51: I _think_ you’ll want to set the `page_title` context variable, so that the `<title>` tag in the `<head>` gets populated?

```python
def get_context_data(self, **kwargs):
    context = super().get_context_data(**kwargs)
    context["page_title"] = "Privacy policy"
    return context
```

zarino (approved):
Looks good. Just one comment about the missing page title.

PR #572, opened 2023-10-04
[Scorecards] Load council type tables over Ajax

### LINKED ISSUE #512: Load 2023 Scorecards tables via Ajax, rather than full page load
The council type switcher buttons (and “find your council in the table” search box) should load the different council tables in place using Ajax, to make things feel a bit snappier.

We’ll want to use `history.pushState` to update the URL (to `/scoring/<council-type>/`).

Separate `/scoring/<council-type>/` should still exist, as a non-JavaScript fallback, and as a landing page that people can link directly to.

zarino:
- JavaScript for intercepting council type button presses, and loading the pages over Ajax instead, pulling out specific elements from the new page, and injecting them into the current page, so that we don’t have to mess with resetting event listeners on any parent elements.
- Update browser URL with History.pushState, and (re-)load pages over Ajax on popState event, so that browser back/forward buttons work as you’d expect.
- “loading shimmer” animation for elements being replaced by Ajax.
- Remove conditional around “Advanced filters” button – no need to hide the Advanced Filters for Combined Authorities or NI Councils.
- Save some vertical space in the viewport by removing the useless “Scorecards results” heading and moving the “Advanced filters” button into a space above the table on the right.

Fixes #512.

https://github.com/mysociety/caps/assets/739624/9b905769-0ef6-4015-9ca4-f9ecde0568ba

### DIFF
--- a/scoring/static/scoring/js/main.js
+++ b/scoring/static/scoring/js/main.js
@@ -364,3 +364,57 @@ forEachElement('.js-methodology-council-autocomplete', function(input){
         container.setAttribute('data-methodology-active-council-type', council.council_type);
     });
 });
+
+function ajaxLoadCouncilTypeScorecard(url) {
+    var selectors = [
+        '#council-type-filter',
+        '#advancedFilter .modal-body',
+        '.table-header h3',
+        '.scorecard-table'
+    ];
+
+    selectors.forEach(function(selector){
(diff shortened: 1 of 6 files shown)


lucascumsille:
Looks great @zarino

lucascumsille (approved):

Issue #545, opened 2023-09-05
Nice to have Scorecards 2023


lucascumsille:
### General

- [x] Make the focus state more obvious
- [x] Navbar hover and focus state more obvious.
- [x] Main navbar with a light gradient
![Screenshot 2023-08-29 at 17-01-32 Council Climate Plan Scorecards](https://github.com/mysociety/caps/assets/13790153/37bbe575-a6fd-468f-a7ed-8c2527a7bd12)

### Homepage

- [ ] Enable mobile view for sections on home table
- [ ] Consider stronger visual “connection” between homepage table jump/filter controls and the table itself

### Council page
- [ ]  When applying filter and there are no results, we probably want to display a message.
- [x]  When the accordion button is active, I should change the colour to make it obvious that the state is different.
- [x]  I left some TODO for Struan, things like if the question is council operational only do this. Or add classes to the row.
- [ ] Add highlights sections. For example "This council is a top scorer in "Biodiversity""
- [ ] Add a political control message for council that change on 2023.

### Section child page

- [ ] Way to travel to another section

zarino:
FYI @lucascumsille here’s something I was playing with earlier, when I was trying to create a clearer visual connection between the homepage council type filters / "jump to council in table" search input, and the table itself…

![Screenshot 2023-09-28 at 08-13-33 Council Climate Plan Scorecards](https://github.com/mysociety/caps/assets/739624/8e83b48a-62d0-475e-96e4-1ec9e2cc403f)

My thought process: Exploring "tabs" as a way to represent the different council type tables (rather than little buttons), and then moving the "jump to council" search box nearer the table itself, near to the advanced filters, since those all feel like actions you take on the table. This left me with the "external" links to the sections page and a council’s individual page, which I—perhaps controversially—ended up moving down to the bottom, alongside a "card" for the (paid-for) spreadsheet download offer.

Would be interesting to discuss with CEUK, see whether they feel this is an improvement over what we have at the moment.

lucascumsille:
I mean I like it, but my instant thought is: are people going to scroll down and see it? I think if we are going to make this changes it would also mean to access all the items below in the navbar. But last time I remember Isaac wasn't keen on putting the section there. 
If we are not including them in the navbar, then adding a content navbar stick to the bottom, potentially one that could be opened or closed could be an approach.

I feel like we are back to the discussion from the vert begining and having a scorecard table page, and leaving the homepage as a way to present the project and main features, but CEUK wasn't too keen for that.

PR #573, opened 2023-10-05
Scorecards 2023 sections comparisons

struan:


scoring/templates/scoring/section.html
```diff
@@ -414,33 +390,28 @@ <h5 class="modal-title" id="section-{{ forloop.parentloop.counter }}-question-{{
...
                       </td>
 
+                      {% for comparison in question.comparisons %}
                       {% comment %} On mobile we will display only the average for the type of council. {% endcomment %}
                       <td data-column="score" class="text-start text-md-end d-flex flex-row d-md-table-cell score">
-                        {% comment %} TODO add score {% endcomment %}
-                          <span class="me-2">1/3</span>
-                          <span class="badge bg-secondary d-md-none">Boerset Council</span>
+                        <span class="me-2">{{ comparison.score }}/{{ question.details.max_score }}</span>
+                        <span class="badge bg-secondary d-md-none">{{ comparison.council }}</span>
```
lucascumsille, line 397: In here is not pulling the name of the council. This is the mobile view just in case.

lucascumsille (changes requested):
Thanks @struan it looks. Just one thing that. I couldn't figure how to fix.

scoring/templates/scoring/section.html
struan: ooops, fixed.

zarino (changes requested):
The layout of this page is such a mess, it’s difficult to tell what’s working and what’s not, and which of those is intentional and which isn’t. I’ve rebased on top of the latest `scorecards-2023` branch, so we _should_ be seeing everything as up to date as it can be.

Can someone confirm whether they see these issues too, or whether it’s just because I need to re-import some data or something?…

- On `/sections/<whatever>/`, once you pick a council type, a box appears between the "Add councils to compare" card and the question table, showing the average score for that type of council. Why is this repeated here, when it was already displayed at the top of the page?
- It seems that I’m able to add _any type_ of council as a comparison, even on the pages for sections that don’t apply to that type of council. For example, on `/sections/s5_bio/` (a section that, by its own admission in the boxes at the top of the page, doesn’t apply to CAs, and doesn’t even have an equivalent for CAs) I was able to add North East Combined Authority as a comparison – and it appeared there, as a column, with scores in it! How is that even possible?
   - Edit: I should add, I think it’s fine if we’re doing some sort of magic to display the scores for the most relevant applicable section – but we need to make that clear to users.
- Similarly, when you visit the page for a section that _only_ applies to CAs (eg: `/sections/s2_tran_ca/`), the page still offers to "show questions by type of council" with four buttons, one for each of the non-CA council types! I think you probably meant to hide those filter buttons on the CA-only sections – or hide the entire pale purple filter form on CA-only sections. The user’s somehow made their way to the page for a CA-only section, we can probably assume they’re only interested in CAs?
- When you add a council to compare, the little button with the council’s name (that appears inside the "Add councils to compare" box) smashes against the "Clear all" link – `<div class="selected-council-wrapper">` needs an `mb-3` or something.

Aside from those, there are a _bunch_ of other things that just look or feel broken, but aren’t strictly related to adding comparison councils, so probably shouldn’t be fixed in this branch. But they fell too important to add to #545. Maybe we need a new snagging ticket 🤔 in the meantime, here they are, better than nothing…

- There is no way to get to a section page (eg: `/sections/s1_b_h/`) or the overall Sections page (`/sections/`) from the homepage. The purple "Go to sections" button at the top centre of the homepage doesn’t work.
- `/sections/` doesn’t currently show a table, by default, on pageload. My instinct is this page should work like the homepage, it should "select" the Single Tier button by default, until the user chooses otherwise.
- We’re obviously missing section descriptions on `/sections/` page.
- It felt not very obvious to me that you could click the entire "cards" on the `/sections/` page. I think because they don’t react when you hover over them, and there’s no call to action (underlined link or button) inside the card to suggest that you even _should_ hover near them.
- `/sections/` and `/sections/<whatever>/` are both missing a page title (browser just displays: `| Council Climate Scorecards`)
- We’re obviously missing text content for the "Section information" accordion on `/sections/<whatever>/`.
- Like `/sections/`, `/sections/<whatever>/` doesn’t currently show a table, by default, on pageload. As above, I think this page should behave like the homepage, with the "Single Tier" button pre-selected (or "Combined Authority" if this is a section that only applies to CAs), until the user chooses otherwise.
- When you "select" a council type button under "show questions by type of council" on `/sections/<whatever>/`, the button doesn’t display as "active".
- Even worse, when you click that council type button, the entire page reloads and you’re left staring at the top of the new page, while the table of questions is (in my case) at least two full page scrolls down below. Yuck.
- The markup for question rows inside the question table on `/sections/<whatever>/` isn’t the same as the markup for question rows on `/councils/<whatever>/`. As a result, the wording for each question on `/sections/<whatever>/` crashes into the "How this is marked" button below, because the question wording is wrapped in a `<div class="js-collapse-children">` (rather than just the `<p>` used for questions on the council page). I’m getting serious [code smell](https://en.wikipedia.org/wiki/Code_smell) from the fact that these two tables don’t appear to be reusing a common template for rendering these rows – or at least the content of this particular cell!

lucascumsille:
- [ ] On /sections/<whatever>/, once you pick a council type, a box appears between the "Add councils to compare" card and the question table, showing the average score for that type of council. Why is this repeated here, when it was already displayed at the top of the page? 
I just added a commit to display that only when a council has been selected. That way users don't need to scroll up to compare their council with their type of council.

Screenshot 2023-10-11 at 06 55 32

- [ ] Similarly, when you visit the page for a section that only applies to CAs (eg: /sections/s2_tran_ca/), the page still offers to "show questions by type of council" with four buttons, one for each of the non-CA council types! I think you probably meant to hide those filter buttons on the CA-only sections – or hide the entire pale purple filter form on CA-only sections. The user’s somehow made their way to the page for a CA-only section, we can probably assume they’re only interested in CAs?
I think it's fine to assume the users would be only interested on CA sections. So I have added a commit to hide the filters when we are in a CA type of council.

- [ ] When you add a council to compare, the little button with the council’s name (that appears inside the "Add councils to compare" box) smashes against the "Clear all" link – <div class="selected-council-wrapper"> needs an mb-3 or something.
Added a commit to fix that.

lucascumsille:
@zarino the rest of the comments I added them to [this ticket.](https://github.com/mysociety/caps/issues/577), because it won't be updating just text, also there is some changes layouts from CEUK. It also includes things like section descriptions on the `/sections` page.

lucascumsille:
Just in case there is one bullet point that still need work:

- [ ] It seems that I’m able to add any type of council as a comparison, even on the pages for sections that don’t apply to that type of council. For example, on /sections/s5_bio/ (a section that, by its own admission in the boxes at the top of the page, doesn’t apply to CAs, and doesn’t even have an equivalent for CAs) I was able to add North East Combined Authority as a comparison – and it appeared there, as a column, with scores in it! How is that even possible?

lucascumsille:
@struan here are the descriptions in the meantime:

Buildings & Heating
Buildings and Heating is one of the biggest sectors of carbon and other greenhouse gas emissions in the UK. This section covers the main actions that councils can take to support both private rented and owned homes and socially renting households to reduce the emissions from their homes. 

Transport
Transport is the other biggest sector of greenhouse gas emissions in the UK. This section covers the main enabling actions councils can take to reduce car use and encourage more sustainable transport within their area. 

Governance & Finance
This section aims to understand to what extent climate action has been incorporated and embedded across the whole of the council in all its activities and services in its decision making, forward planning and structures. This section also looks at how councils are raising funds for climate action and whether the councils’ investments are sustainable or supporting high carbon infrastructure and industries.

Planning and Land Use
This section focuses primarily on how councils are using their planning powers, primarily through their Local Plans, to ensure low emission new buildings and homes, as well as ensuring new developments are built to minimise their environmental impact. This section also covers the renewable energy generation and fossil fuel generation planning applications in the area.

Biodiversity
The climate emergency is deeply connected to the ecological emergency. This section looks at what councils can do to protect and increase biodiversity in the area through their direct actions, the management of their green spaces, and biodiversity net gain requirements for developers. 

Collaboration & Engagement
This section addresses how councils can collaborate with others to improve their own climate action and to support others in the area to decarbonise. More than half of the emissions cuts needed to reach net zero rely on people and businesses taking up low-carbon solutions, and councils can work with those in their local area to enable those solutions.

Waste Reduction & Food
This section looks at the influencing role councils can play in supporting sustainable food production on their land and in their schools, and circular economy initiatives locally. Councils also have an important role to play in waste and recycling locally and improving this. 

### For Combined Authorities
Buildings, Heating & Green Skills
Buildings and Heating is one of the biggest sectors of carbon and other greenhouse gas emissions in the UK. Combined Authorities do not own council homes or have direct involvement with homeowners. Instead, they lead on adult education in their area, as well as economic growth so have an opportunity to influence the Green Skills needed for low carbon buildings and heating. 

Transport
Transport is the other biggest sector of greenhouse gas emissions in the UK. This section covers the main actions combined authorities can take at a broader level to reduce car use and encourage more sustainable transport within their area.

Governance & Finance
This section aims to understand to what extent climate action has been incorporated and embedded across the whole of the combined authority in all its activities and services in its decision making, forward planning and structures. This section also looks at how combined authorities are raising funds for climate action.

Planning, Biodiversity & Land Use

The climate emergency is deeply connected to the ecological emergency. This section covers how combined authorities can support the protection of and increased biodiversity in the area. This section also covers the enabling role combined authorities can play in supporting renewable energy generation locally and embedding net-zero targets into their area-wide strategic planning.

Collaboration & Engagement
This section addresses how combined authorities can collaborate with others to improve their own climate action and to support others in the area to decarbonise. More than half of the emissions cuts needed to reach net zero rely on people and businesses taking up low-carbon solutions, and combined authorities can work with those in their local area to enable those solutions.

zarino (approved):
This looks ok to merge :+1:

PR #574, opened 2023-10-05
Update homepage for 2023 categories etc

struan:


lucascumsille:
Looks good to me. I added a fix when the council type is too long:

Screenshot 2023-10-05 at 14 42 42

Also not sure if it's part of this PR or maybe I'm the only one having the problem, but when I try to use the advanced filters I get this error:

Screenshot 2023-10-05 at 14 53 13

I tried to use the "Country" filter

struan:
Cool, I've fixed the country filter problem.

lucascumsille:
Thanks @struan 
Not sure if it's my local, but testing the filter I came up with these results when applying "England" and & "Rural" in the advanced filter. Basically it's showing Aberdeen city council as my first result, but that should be in Scotland.
Screenshot 2023-10-09 at 06 44 56

struan:
Fixed the filter issue

lucascumsille (approved):

PR #575, opened 2023-10-05
Pull scorecards 2023 methodology from db

struan:


struan:
You will need to a) fetch updated question import data for this and b) re-run the import_actions_scores command.

lucascumsille (approved):

PR #576, opened 2023-10-09
Methodology page

lucascumsille:
Fixes: #541 
Updated methodology page copy. According to this document:
https://docs.google.com/document/d/1W6onpYh5B2Br7v6PUTV8OUpRu-ehNc3R24DZaTUKOqg/edit

Preview here:

https://github.com/mysociety/caps/assets/13790153/556cce6e-41de-47af-ab07-4991b3b5118d

PR #580, opened 2023-10-11
store section descriptions in the database and use them on section pages

struan:


zarino (approved):
Looks good!

I note https://github.com/mysociety/caps/pull/576/files includes what I assume are hard-coded versions of some of these section descriptions, in the Methodology page. I wonder whether we should merge that PR and rebase this branch on top so those descriptions can come out of the database, or whether we should merge this one first and then amend the Methodology PR separately?

struan:
I would like to get the sections branch merged if it's fine and then then this branch and then we can fix the methodology branch after.

PR #581, opened 2023-10-11
generate 404 page instead of 500 when section or score not found

struan:
tidies up a few loose ends of places where we're not handling missing objects correctly.


zarino (approved):
👍

PR #585, opened 2023-10-12
Switch to cookieless approach to scorecards

ajparsons:
Switch from old style (doesn't work) to new style (does work) approach to running google analytics without cookies.


zarino:
Thanks @ajparsons, but it turns out we already fixed this (for both the 2022 and 2023 Scorecards sites) in the upcoming `scorecards-2023` branch – https://github.com/mysociety/caps/commit/222d73ae15742602e29ccf7cf6466c04f23cbdc4 "Force Google to use anonymous analytics"

Rather than merging this now, and adding just another thing to rebase out during the merge of `scorecards-2023`, I figure we’re probably fine to leave the existing tracking code in place on the 2022 site, until the Action Scorecards launch on Wednesday morning.

PR #598, opened 2023-10-23
[scorecards] fix the show specific council section page dropdown

### LINKED ISSUE #594: [scorecards] show questions for a council dropdown on section pages not working
This should be a link to `/sections/$code/?council=$slug` but it does nothing.

struan:
Fixes #594

### DIFF
--- a/scoring/static/scoring/js/main.js
+++ b/scoring/static/scoring/js/main.js
@@ -341,6 +341,27 @@ forEachElement('.js-methodology-council-autocomplete', function(input){
     });
 });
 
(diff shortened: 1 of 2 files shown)


zarino (approved):
This isn’t _actually_ doing what the search box is meant to do (it’s meant to select one of the council type buttons below, based on the `council_type` of the council you select in the autocomplete) – but I guess adding the council as a comparison is fine, instead, since it automatically selects the right council type and also adds a column with the council’s mark on each question, which people might find useful.

struan:
Oh right, that is not at all what I would have expected it to do.

PR #601, opened 2023-10-24
fix missing section page question criteria

struan:
Also tiny bit or de-deduplication refactoring.


zarino (approved):
Can confirm criteria now appear in the modal on Section page, when they didn’t previously :+1:

Issue #623, opened 2024-01-29
"Plan Scorecards" still appears in <title> for council pages on the Action Scorecards site


zarino:
Page titles for council pages should read, eg:

> Borsetshire County Council Climate Action Scorecard | Council Climate Scorecards

Instead of the current:

> Borsetshire County Council Climate *Plan Scorecards* | Council Climate Scorecards

When you fix this, double check that it affects _both_ the `<title>` and the `<meta property="og:title">` tags.

PR #626, opened 2024-02-06
[Scorecards] Fixed council title property

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/623

Title element and property="og:title" were still using "plan" for the Climate Action Scorecards. Both elements have been updated to replace "Climate Plan Scorecards" with "Climate Action Scorecard"


zarino (approved):
Thanks!

PR #639, opened 2024-03-12
Improve wording of council documents card "Info" header

zarino:
Also tighten up the vertical spacing, while I was in the area.

Screenshot 2024-03-12 at 16 17 28


struan (approved):
🗒️

PR #669, opened 2024-05-29
Draft Methodology 2025 feedback

lucascumsille:
This PR fixes row 22 and 23:
https://docs.google.com/spreadsheets/d/19W5anmr3VIgNmR9uN--SV6fC96AIwYx29WGnbJyb6Eo/edit?pli=1#gid=63266911

- Orange navigation content bar
- Updated pdf version for the methodology

### DIFF
--- a/scoring/static/scoring/scss/main.scss
+++ b/scoring/static/scoring/scss/main.scss
@@ -64,3 +64,4 @@
 @import "toggle-buttons";
 @import "loading-shimmer";
 @import "open-graph-preview";
+@import "years";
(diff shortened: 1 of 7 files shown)


zarino (approved):
I’ve pushed a few fixup commits as we discussed last week – using the position of the nav, rather than its background colour, for emphasis, and switching to purple for the nav link colours.

I‘ve also added a commit that overrides `.bg-primary-*` styles inside a `.year-2025` container, which feels like it might become more and more useful as we start adding more 2025 pages to the site. Eventually, I guess, we’ll replace the `*-primary2025` styles entirely, and rely just on this `.year-2025` inheritance.

Issue #665, opened 2024-05-21
[Scorecards] Add data visualisations to Section pages


zarino:
We’re creating one-page visualisations (roughly 16:9 landscape) for about 10 of the 2023 Scorecards sections.

It’d be nice to embed those visualisations on the relevant Section pages, eg: https://councilclimatescorecards.uk/sections/s4_g_f/

We could also set them as the `og:image` on those pages, so they appear when the page is shared!

PR #677, opened 2024-07-10
Added Social Media Graphics to Section pages

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/665

### Desktop

https://github.com/mysociety/caps/assets/13790153/23f3d860-f20d-4a15-b550-b14a8a717a4c

### Mobile

https://github.com/mysociety/caps/assets/13790153/ef650f9f-7854-4176-be6b-4fa1b4da6ebc

### Nav button for Mobile
Let me know what you think of this one. I think there is quite a lot of scrolling on mobile, so this might not be a bad idea.

https://github.com/mysociety/caps/assets/13790153/24c896ab-0ff3-4ab5-80af-78e95ea29c6a


zarino (approved):
Generally looks good!

I’ve made a few changes, but will likely have squashed them into your commits by the time you read this. So for posterity:

I wasn’t sure we needed to duplicate the loop through the graphics in the template (creating a separate sideways-scrolling gallery on desktops, and a vertically stacked gallery on mobiles) so I’ve added a commit that combines those – it’s now a horizontally scrolling gallery on all screen sizes, but I loved how you switched between square and wide images depending on viewport size, so I’ve kept that!

I’ve also added JPG previews of the PDFs, and put them at the end of the gallery – they’re such pretty infographics, it seems a shame to hide them behind a button that says just "See PDF version"! And, I wanted to generate JPG versions anyway, to use as the OpenGraph image for the page, which I’ve also now added.

I noticed Git was messing with the LF/CRLF line endings in the PDFs (and—perhaps coincidentally, perhaps because of this—when I checked out your branch, all the PDFs were corrupted). So I’ve re-exported and re-added the PDFs from Figma, and marked PDFs as "binary" in `.gitattributes` so Git doesn’t mess with them again. Bonkers that Git treats PDFs as text files by default.

Finally, there was a typo of "governanace" in some of the image file paths, so I fixed that.

I’ll go ahead and squash, merge, and deploy this :+1:

PR #682, opened 2024-07-25
[Scorecards] Added Building and Heating Graphics

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/681

Adds social media graphics to Building and Heating section.

I added some conditional, because it's the first time we have uneven number of graphics, in this case Facebook has one extra in comparison to Instagram.


zarino (approved):
Fab. I’ve tidied up the indentation on those conditionals blocks you added, and also included two other small changes that CE UK requested, as separate commits.

PR #683, opened 2024-07-29
add a documents end point to the API

struan:
Lets people get the list of council documents from the API


zarino (approved):
Works locally, and looks good to me :+1:

Issue #688, opened 2024-09-03
Track clicks on section page infographic / social graphic download buttons


zarino:
We’d like some indication of whether people are using the infographics and social graphics, in the gallery on section pages.

PR #692, opened 2024-09-11
[Scorecards] Added click tracking for graphics

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/688

We are tracking:

- Section
- File type
- if user is clicking an image then we also track which ratio

### DIFF
--- a/scoring/static/scoring/js/main.js
+++ b/scoring/static/scoring/js/main.js
@@ -436,33 +436,28 @@ if ( typeof window.fetch !== 'undefined' ) {
     });
 }
 
-// Graphic click tracking
+// Graphics click tracking
 var trackEvent = function(eventName, params, callback){
     params = params || {};
     callback = callback || function(){};
     params['event_callback'] = callback;
     setTimeout(callback, 2000);
+    console.log(params)
     gtag('event', eventName, params);
-};
-
-// Function to extract section from the URL
-var getSectionFromURL = function() {
-    var urlParts = window.location.pathname.split('/');
-    return urlParts[urlParts.length - 2];
 };
 
 document.querySelectorAll('.js-social-graphic-download').forEach(function(el) {
     el.addEventListener('click', function(e) {
-        e.preventDefault();  // Prevent default page navigation
-
+        e.preventDefault();
         var eventName = "download";
         var params = {
-            'file_type': el.getAttribute('fileType'),
-            'section': getSectionFromURL()
+            file_type: el.getAttribute('file-type'),
+            // The url will contain the section
+            section: window.location.href
         };
 
         // Only add image_ratio if it exists
-        var imageRatio = el.getAttribute('imgRatio');
+        var imageRatio = el.getAttribute('image-ratio');
         if (imageRatio) {
             params['image_ratio'] = imageRatio;
         }
--- a/scoring/static/scoring/js/main.js
+++ b/scoring/static/scoring/js/main.js
@@ -435,3 +435,40 @@ if ( typeof window.fetch !== 'undefined' ) {
         }
     });
 }
+
+// Graphic click tracking
+var trackEvent = function(eventName, params, callback){
+    params = params || {};
+    callback = callback || function(){};
+    params['event_callback'] = callback;
+    setTimeout(callback, 2000);
+    gtag('event', eventName, params);
+};
+
+// Function to extract section from the URL
+var getSectionFromURL = function() {
+    var urlParts = window.location.pathname.split('/');
+    return urlParts[urlParts.length - 2];
+};
+
+document.querySelectorAll('.js-social-graphic-download').forEach(function(el) {
+    el.addEventListener('click', function(e) {
+        e.preventDefault();  // Prevent default page navigation
+
+        var eventName = "download";
+        var params = {
+            'file_type': el.getAttribute('fileType'),
+            'section': getSectionFromURL()
+        };
+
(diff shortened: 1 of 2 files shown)


lucascumsille:
@zarino thanks for all the help when creating the issue. I'm not sure how to test it though, I added some console logs before to check if we were gathering correctly the params, event name and callbacks, and it seems to be working fine.

I split the fileType and imgRatio, but we could have merged those two together, I think because they are two separate things it makes sense to keep them separate, but I'm not to fuzzy about either option.

scoring/static/scoring/js/main.js
```diff
@@ -435,3 +435,40 @@ if ( typeof window.fetch !== 'undefined' ) {
...
+    return urlParts[urlParts.length - 2];
+};
+
+document.querySelectorAll('.js-social-graphic-download').forEach(function(el) {
+    el.addEventListener('click', function(e) {
+        e.preventDefault();  // Prevent default page navigation
+
+        var eventName = "download";
+        var params = {
+            'file_type': el.getAttribute('fileType'),
```
zarino, line 460: A small thing, but no need for quotes around object keys in JavaScript.

scoring/static/scoring/js/main.js
```diff
@@ -435,3 +435,40 @@ if ( typeof window.fetch !== 'undefined' ) {
...
+    params = params || {};
+    callback = callback || function(){};
+    params['event_callback'] = callback;
+    setTimeout(callback, 2000);
+    gtag('event', eventName, params);
+};
+
+// Function to extract section from the URL
+var getSectionFromURL = function() {
+    var urlParts = window.location.pathname.split('/');
```
zarino, line 450: I’m slightly worried this is a bit fragile if we happened to change the URL pattern, and some missing data in GA would go easily unnoticed. Since we’re pulling lots of other data out of custom attributes on the `<a>` elements, maybe we should do the same here?

scoring/templates/scoring/section.html
```diff
@@ -133,7 +133,7 @@ <h3 class="fs-4 mb-0 text-center">
     <div class="mb-2 d-flex align-items-center justify-content-between">
         <h3 class="mb-0">Key takeaways</h3>
       {% if social_graphics.zip %}
-        <a class="btn btn-sm btn-outline-primary" href="{% static social_graphics.zip %}">Download all (.zip)</a>
+        <a class="btn btn-sm btn-outline-primary js-social-graphic-download" fileType="zip" href="{% static social_graphics.zip %}">Download all (.zip)</a>
```
zarino, line 136: Another small thing, but it’d be neater (and keep the HTML5 valid) to use [data attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) for those file type and image ratio properties on the links – so `data-file-type` and `data-img-ratio`. You can then pull them out again in the JavaScript with `el.dataset.fileType` and `el.dataset.imgRatio`.

zarino (changes requested):
Thanks @lucascumsille! A couple of small comments inline.

Although—crazy thought—I wonder whether we _even need_ to be recording `section`, `fileType` and `imgRatio` separately, when we _could_ just track the URL of the image the user is downloading? 🤔 Wonder whether that might be simpler.

If you want to test this out, you _should_ theoretically be able to do it by putting a valid Google Analytics Gtag ID (doesn’t matter which, can be a live account, GA will ignore the events because they’re not coming from the right domain) into a `GOOGLE_ANALYTICS_SCORECARDS` environment variable in your `.env` file, and then using [a browser extension like this](https://chromewebstore.google.com/detail/taghound-analyticsgtmpixe/canpneabbfipaelecfibpmmjbdkiaolf) to intercept/log the `gtag` calls (or, without an extension, you’ll still be able to see the calls being made, as Network requests, in your built-in browser tools).

lucascumsille:
Thanks for the feedback @zarino 🙌

Regarding the following comments:
>Although—crazy thought—I wonder whether we even need to be recording section, fileType and imgRatio separately, when we could just track the URL of the image the user is downloading? 🤔 Wonder whether that might be simpler.

I thought we were making the split between the different properties because it had an impact for the person checking the clicks, but if it doesn't have any practical benefit then I would go for the simple solution. Let me know if you have any preference regarding this.

>I’m slightly worried this is a bit fragile if we happened to change the URL pattern, and some missing data in GA would go easily unnoticed. Since we’re pulling lots of other data out of custom attributes on the <a> elements, maybe we should do the same here?

Good point! I tried finding how we pull the section codes on a section page, I couldn't find anything there, except by the `canonical_url` bit. I'm sure we can add something to `sectionView` if we wanted to add it as an attribute. In the meantime I have replace the section by the current page url, which gives up the section code.

zarino:
> Let me know if you have any preference regarding this.

My instinct is simpler is better. There won’t be a huge number of events here, and if somebody really wanted to split out the filetypes and aspect ratios, they could do it from the image URLs.

But maybe check that just recording the URLs of the images, and how often they’ve been downloaded, is good enough for CEUK? Since they’re the ones that asked for the tracking in the first place :-)

lucascumsille:
@zarino at the end we will be using only the url as parameter, considering that it also displays:

- section
- File type
- Image ratio

zarino (approved):
👍 🎉

lucascumsille:
Closed via https://github.com/mysociety/caps/commit/97b851cec6e2f129a491bc146475413e314e1cec

Issue #686, opened 2024-08-20
[scorecards] look at including the nearest neighbour tool on council page


struan:
To make deciding on which council's to compare a bit easier.

PR #698, opened 2024-10-14
Added council suggestions for comparison tool

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/686

I left some comments regarding things that are pending, but front-end related I think it should be ready.

- [x] We need to pull a list of 5 councils based on overall similarity like we do in CAPE. I picked 5 randomly, it seemed like a sweet spot between giving the user enough options, but not overwhelm them with too many choices.
- [x] Replace the placeholder councils with a loop based on overall similarity.
- [x] We need to add commas to the list of councils, but I can do that using CSS, however I thought we could keep doing it with Django like we have done in other part of the website. Let me know which approach you prefer.

### Preview here:

Screenshot 2024-10-21 at 14 19 38

Screenshot 2024-10-21 at 14 19 12


lucascumsille:
@zarino and @struan I think I'm done with this PR, but we need the back-end logic to be implemented. I left some comments at the beginning.

zarino:
@lucascumsille I’ve wired this up, so that a maximum of 5 overall related councils are passed to the template, removing any that are already selected for comparison. I simplified some of the wording, and used flex layout for the links, rather than concatenating them into a sentence with commas.

Screenshot 2024-11-06 at 16 03 28

zarino:
Marking as ready for review by @struan, so he can check I my Python isn’t too horrendous, and work out whether we should fix the linting errors or not.

scoring/views.py
```diff
@@ -407,28 +407,41 @@ def get_context_data(self, **kwargs):
...
+        context["comparisons"] = comparisons
+
+        for group in council.get_related_councils(5):
+            if group["type"].slug == "composite":
+                if comparisons:
+                    # Filter out any related councils that are already being compared
+                    context["similar_councils"] = [
+                        sc
+                        for sc in group["councils"]
+                        if sc.slug not in {score.council.slug for score in comparisons}
```
struan: fwiw I've pushed a small fixup so we're not regenerating this set every time

scoring/views.py
```diff
@@ -407,28 +407,41 @@ def get_context_data(self, **kwargs):
         context["sections"] = sorted(
             sections.values(), key=lambda section: section["code"]
         )
+        context["comparisons"] = comparisons
+
+        for group in council.get_related_councils(5):
+            if group["type"].slug == "composite":
```
struan, line 413: why does this only apply if it's composite?

struan (approved):
this looks good.

I've pushed some commits to update the black version and fix the resulting linting issues. We were using an old version so I've updated it to the latest version.

scoring/views.py
zarino: We only want the 5 "overall" similar councils, not the other types of similarity.

zarino: (I just copied this block, wholesale, from the caps app)

PR #702, opened 2024-11-01
Avoid displaying ".0" decimals on question marks

### LINKED ISSUE #699: [Scorecards] Integer numbers instead of float numbers for question marks
Minor issue raised by CEUK, but in some question we have marks that are displayed as float numbers instead of integer, which seems a bit odd when reading the whole table. 

There are questions where using float numbers make sense because they can be eg. 1.25, we wouldn't want to lose that accuracy when it's necessary. Could we add in Django a conditional to only display float numbers if needed and display integer by default?

Screenshot 2024-10-21 at 11 03 28

Example extracted from:
https://councilclimatescorecards.uk/councils/london-borough-of-islington/

zarino:
Fixes #699. The question, as outlined there, is:

> Can the number of marks a council got on a question ever be a fraction with more than one decimal place? And if it can, do we care if it gets rounded to one decimal place?

If the answer to either of those questions is "no" then we should be good to merge this PR.

### DIFF
--- a/caps/templatetags/caps_templatetags.py
+++ b/caps/templatetags/caps_templatetags.py
@@ -71,6 +71,22 @@ def percentage(value: float):
     return f"{value * 100:.0f}%"
 
 
+@register.filter
+def format_mark(value):
+    """
+    Display whole-number marks as a whole numbers, and
+    fractional marks as decimals, removing trailing zeroes.
+    """
+    try:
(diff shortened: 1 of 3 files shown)


lucascumsille:
Thanks @zarino I'm checking with CEUK if that would be a problem.

PR #707, opened 2024-11-15
Split single tier types

### LINKED ISSUE #611: [scorecards] update filtering to enable spliting metro and non metro districts

zarino:
Fixes #611.

Adds an "Advanced filter" for `authority_type`, on the Single Tier scorecard.

Screenshot 2024-11-15 at 18 50 39

While I was there, I also fixed a long-standing code smell, where we had been using `authority_type` to refer to both _actual_ MapIt council authority types (eg: "UTA", "DIS", "CTY") and Scorecards scoring groups (eg: "single-tier").

This makes the PR look much more complex than it actually is. Best to review each commit separately!

### DIFF
--- a/caps/models.py
+++ b/caps/models.py
@@ -480,20 +480,10 @@ class Council(models.Model):
         return df
 
     def get_scoring_group(self):
-        if self.country == self.NORTHERN_IRELAND:
-            group = "northern-ireland"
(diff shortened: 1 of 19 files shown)


struan (approved):
Few small fixups added and one additional fix I noticed while in here but yup, looks good.

Issue #644, opened 2024-04-02
[scorecards] Show comparisons between 2023 and 2025 Scorecards scores


zarino:
TBC:

- Do we want to show comparisons for questions, sections, and overall score?
- Do we want to show comparisons on the "homepage" tables, or just the council pages?
- How do we want to handle newly merged councils? Council types might change too. Perhaps best to just avoid trying to "compare" newly formed councils.

Need to remember that questions will be changing between generations – will need a look-up for "what is the relevant 2023 question to compare to?"

lucascumsille:
[In this Slack thread ](https://mysociety.slack.com/archives/CHL40EKB9/p1730276983264899) it might be a good idea to display:

- Score for current year
- Absolute variation
- There is no need to specify "more/less than last year". If we include it might be as a tooltip for desktop.

zarino:
Going to close this, as everything outstanding is now covered in other tickets (like #731) or pull requests (like #735).

PR #729, opened 2025-02-13
Year comparison Council and Section Tables

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/644

### Preview Default

https://github.com/user-attachments/assets/a50cb3f9-1854-4345-9c46-885bc2276c55

### Preview Comparison

https://github.com/user-attachments/assets/eceeb597-5f0e-46a4-916c-353db30ce40e

## Preview Mobile

https://github.com/user-attachments/assets/8f254af0-8a1d-4f09-9e18-e4fb58e6203d


lucascumsille:
@zarino I haven't changed the section table, I'd like to see if you are happy with this first.

scoring/static/scoring/scss/table-question-council.scss
```diff
@@ -72,3 +72,33 @@ $question-types: (
         }
     }
 }
+
+// This class helps us avoid using Bootstrap display classes to hide/display elements, especially in tables where JS classes that handle display properties might conflict with the !important rules Bootstrap uses. With this class we wouldn't need to use d-none d-md-table-cell or d-md-block.
```
zarino, line 76: FWIW, another way you could do this is by telling Bootstrap 5 to generate utility classes like `d-lg-revert`, which you could then use in the HTML.

You can do this by adding overriding the "display" section of the big `$utilities` map, the same way we add custom colours, font-weights, and widths in `scoring/static/scoring/scss/_utils.scss`. Eg: 

```scss
"display": map-merge(
    map-get($utilities, "display"),
    (
        values: inline inline-block block grid table table-row table-cell flex inline-flex none revert
    )
),
```

Just a thought!

scoring/templates/scoring/council.html
```diff
@@ -226,77 +243,163 @@ <h3 class="exclamation text-white">Visit us again on a bigger screen</h3>
...
+                <td class="js-previous-year-score display-only-large-up">{{ section.weighted_score|floatformat:0 }}%</td>
+
+                <td class="display-only-large-up">
+                    {% comment %} TODO: IF Negative variation: Change bg-green-100 to bg-red-100 {% endcomment %}
+                    {% comment %} TODO: IF Negative variation: Change text-success to text-danger  {% endcomment %}
+                    {% comment %} TODO: IF Negative variation: Change &#8593; to &#8595; {% endcomment %}
+                    {% comment %} TODO: IF No variation: Change bg-green-100 to bg-gray-300 {% endcomment %}
+                    {% comment %} TODO: IF No variation: Change text-success to text-dark {% endcomment %}
+                    {% comment %} TODO: IF No variation: display '--' {% endcomment %}
+                    <span class="badge align-text-bottom bg-green-100 text-success text-center">&#8593; 5%</span>
```
zarino, line 364: I wonder whether this built-in typographical arrow is just a bit weedy? It’s displayed quite small, and I worry a little bit that people might mistake it for a "1".

We could make a up/down triangle out of CSS borders if we wanted. Or include a simple, chunkier arrow as an SVG or a background image (and could even rotate it with CSS to save having two versions).

zarino (approved):
Seems ok to me @lucascumsille! I’ve left a few comments, but feel free to ignore them if you don’t agree.

scoring/static/scoring/scss/table-question-council.scss
lucascumsille: Yes, the problem with this approach and I why ended up creating that is because in some cases, when using the 2023 score switch the rules ended up conflicting when and element was also using classes like `js-previous-year-score`, because Bootstrap uses `!important` then some element that shouldn't be displayed were still being displayed. As an alternative I thought I could have added the bootstrap classes and add specific rules `js-previous-year-score-display d-lg-table-row` situations.

scoring/templates/scoring/council.html
lucascumsille: Happy to add the SVG ones, I think it's simpler and follows the concept of what we have been doing.

lucascumsille:
@struan I just finished pushing commit to this branch. The Section table now also has the same layout as the council table. Let me know if you'd like to catch up =)

struan:
Merged into 2025-scorecards

PR #735, opened 2025-03-06
[Scorecards] Added year comparison question page

lucascumsille:
- [ ] Confirm what will do with questions that have change their maximum scores between 2023 and 2025. My guess is we should't show any difference. Instead let's display a message below the overview cards explaining why there is no comparison.

### Desktop
https://github.com/user-attachments/assets/b302ed73-6947-49eb-8954-59f9f4b07620

### Mobile
https://github.com/user-attachments/assets/f92d3c56-970d-416b-b967-987788db55b2


lucascumsille:
@struan I think we are ready with the question page, but if you prefer we could wait for tomorrow, however this is the same version CEUK gave the approval last week.
I need to double check with them tomorrow how we will handle when a question has changed its maximum score from one year to another and also when is a new question. At the moment I have added two message that should handle both scenarios, but I'll check with them tomorrow. I think in worst case scenario the copy will change.

lucascumsille:
@struan, in our call with CEUK today, they suggested there might be questions where they wouldn't want a comparison between years, and it won't be a matter of "New question". Some questions changed their criteria quite a bit, and they probably don't want those questions to be compared, so they were asking if there is a way to tackle cases like that.

Regarding the scenario where the "maximum score" changes between years, they haven't decided whether to use the grey message advising users about the score changes. I'll check with them again.

Besides the above, we can move ahead with this branch.

Issue #736, opened 2025-03-11
[Scorecards] Create a Nation page


lucascumsille:
Create a template to display information for each one of the Nations.

### Front-end

- [x] Include it in the navbar
- [x] Section where we include the infographics (How we do it in the section page)
- [x] Table with the scores (How we do it in the homepage)
- [x] For the "Advanced filters" get rid of the nation
- [x] Do we want a "Nations" with stats about all the nations. Similar version of the sections page(`/sections`)
- [x] When users press any of the nations then a message should appear advising them about the new nations page.
- [x] Add metadata like population.
- [x] The "Key takeaways" section won't be updated till after launch, so for now we will hide it.
- [x] Breadcrumbs
- [x] Add a message advising that infographics are coming soon.
- [x] Add description of the nations to the hero section. Use the description we use on the infographics.
- [ ] We need to implement this for 2023 and 2025

### Struan

- [ ] I figure there might be a better way to handle `class NationListView(TemplateView)` and `NationDetailView(TemplateView)`

### After lunch
- [ ] There should be a section with a "Wealth analysis". This table should list all councils + one for total score + 3 columns with KPIs
- [ ] Include infographic
- [ ] Assume there will be year comparison.

lucascumsille:
@zarino I implemented the feedback provided by CEUK yesterday. Not sure if you want me to tell Struan to review this or leave it to you in the meantime.

The only thing that is outstanding creating a version for 2023, but I figure we first implement `class NationListView(TemplateView)`and `NationDetailView(TemplateView)` properly. Not sure if we will keep those classes as they are at the moment.


<img width="1355" alt="Image" src="https://github.com/user-attachments/assets/05ca1a06-7bc9-4fac-be35-00525796701e" />

<img width="1355" alt="Image" src="https://github.com/user-attachments/assets/84a91773-e8d8-4a62-b10c-932619fb7f21" />

zarino:
Thanks @lucascumsille I think @struan and I can take it from here :+1:

PR #746, opened 2025-05-08
Nation pages

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/736


zarino (commented):
@struan The front-end for this is now good to go, but there’s some back-end work required from you next:

- Passing the correct average scores to the template, inside `NationListView`
- Passing the right data to the template from `NationDetailView` to make the table work just like a homepage table

And if we have time:

- Making `NationListView` and `NationDetailView` work for the 2023 Scorecards, as well as the 2025s.

struan:
@zarino I think this is now done but the council type selection on the England page is very much on the functional side so might want some design time.

Issue #751, opened 2025-05-29
Scorecards methodology / how to use pages


lucascumsille:
Apologies for a fairly tight turn-around on these – CEUK have provided finalised text content (below) and we ideally want to have the finished site ready for them to review between 2nd–5th June.

So if you could pencil this into the final few days of Sprint 10, that’d be great.

Text to update with:

- [How To Use page text](https://docs.google.com/document/d/1HzFf3kCHWPrUtilfJEJP0l7K-w92LGUL901o9PSI3F4/edit)
- [Methodology page text](https://docs.google.com/document/d/1eVjUWfw1FOqhYKGLDP4-pIpBiwhxlCmO55sHejAdWuI/edit) – @struan will re-import the [questions and criteria from the spreadsheet](https://docs.google.com/spreadsheets/d/1xu3saUcFxIiOuZ98Nf03Ah8BtFIQYC-S0WqxhVstq2U/edit?gid=64536200#gid=64536200) separately
- [Homepage, council page, and general website text](https://docs.google.com/document/d/1W80lpWuu_dQZ6_BlFlADhqvopd0Ps3Du8s_qgkvU238/edit)

(More files here – https://drive.google.com/drive/folders/1L2-pbZHoEGYcWrcHtBBFxAQAaFSyUhxj – although they should all already be linked from the above three documents.)

PR #752, opened 2025-05-29
2025 scorecards how to use scorecards and methodology

lucascumsille:
Fix: https://github.com/mysociety/caps/issues/751

- [x] Update screenshot in the `How to use scorecards` page. I think in the file [Homepage, council page, and general website text](https://docs.google.com/document/d/1W80lpWuu_dQZ6_BlFlADhqvopd0Ps3Du8s_qgkvU238/edit) we are adding new socials.


zarino (approved):
Looks good @lucascumsille.

I spotted a few changes you missed from the Methodology and How To Use docs (mostly updated links – easy to miss), so have added fixups with those.

And I’ve also added all the other changes that CE UK requested on the other pages, and have taken a screenshot of the new social sharing buttons and added that to the How To Use page, as you spotted.

Just noting, for completeness, that we’re missing:

- URL for the PDF version of the methodology in `templates/scoring/methodology/2025/_intro.html` – waiting on CE UK for that
- `{% url %}` tags for the Nation page links at the top of `templates/scoring/home.html` – need #746 to be merged before we can add these

I’ve marked these up with `#TODO` in the templates, so we can spot them in future.

Issue #764, opened 2025-06-18
[Scorecards] Allow sorting on change in score for overall and section scores


struan:
At the moment you can sort by scores, but not by improvements which would be useful

PR #767, opened 2025-06-26
Allow sorting versus(change) column on home page

lucascumsille:
Fixes: #764 

It also fixes the following bug(note the 'Versus 2023' average cell):

https://github.com/user-attachments/assets/e94407e2-f6b6-4e06-ba0a-5dee5e0097b5

------
### Implementation notes:

- Improvements to the button and .btn element focus state, so it doesn't share the same style as the links.
- Optional: The last commit makes the sorting button wrap the text so the space for clicking is bigger, instead of just having the icon being a button. We couldn't do this before because the names for categories are links, but now that we have moved the sorting to the row below it I figured it would be more UX friendly approach.

### Preview here:

https://github.com/user-attachments/assets/0a5d8282-b049-46e0-aefd-1119c9f190af

- [ ] Pending implement on CA home


zarino (approved):
@lucascumsille Thanks for this. I started making a few small fixups on Friday, before realising that getting the sorting to work on mobile devices would be a whole nother rabbit hole. Anyway, while implementing that I also took the opportunity to fix a load of annoyances with these scorecards-tables, and hopefully make them easier to work with in future.

Do you want to give this branch a quick test yourself, and check you’re happy with how it looks / works, before we merge and deploy?

(It’d be nice to get this deployed on Monday, before the LGA Conference on Tuesday!)

Issue #770, opened 2025-07-16
[Scorecards 2023] Bug on Council page


lucascumsille:
When adding a council to compare on a council page, the added council doesn't have two `td` elements like the main council, this is causing the last column(Council with full marks per question) to be displayed in the same column as the compared council one.

You can replicate here: https://councilclimatescorecards.uk/2023/councils/westminster-city-council/?comparisons=london-borough-of-camden#questions

https://github.com/user-attachments/assets/dbf79971-7a17-4fbd-913f-070ad955ed56

PR #771, opened 2025-07-16
[Scorecards] Added fix to council page 2023

lucascumsille:
Fixes: https://github.com/mysociety/caps/issues/770

When adding a council to compare the columns were mismatched. With this commit if the compared council doesn't have previous year information, then the `td` for the score will spawn two columns, instead of one.


zarino (approved):
Looks good to me!

PR #784, opened 2025-12-04
Filter councils/scores on Question pages by nation/region/imd/etc

### LINKED ISSUE #769: [Scorecards] Add filters to question page
Same filters as elsewhere but for question data

zarino:
This mostly works, but I wasn’t sure what to do about the "Points distribution" statistics, since they’re compared across years, and the filter data is year-specific. I’ve had a stab at a potential solution, but TBH I barely understand what the code is doing at that point, so I’m not confident about it. @struan do you have any thoughts?

Screenshot 2025-12-04 at 17 45 29

Fixes #769.

### DIFF
--- a/scoring/models.py
+++ b/scoring/models.py
@@ -717,6 +717,19 @@ class PlanQuestion(models.Model):
 
         return counts
 
+    def get_score_breakdown_for_councils(self, council_ids, year):
+        counts = (
+            PlanQuestionScore.objects.filter(
+                plan_score__year=year,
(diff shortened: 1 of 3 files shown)


struan:
my concern with the comparing the filtered list from this year with last is that if you then look at the same filter for the previous year you are going to get different numbers so I guess we possibly want to include something to say that it is only comparing for the councils matched this year, rather than all councils that match in the previous year?

PR #786, opened 2026-01-02
Update python, poetry and some dependencies

sagepe:
Bump to Python 3.11, Poetry 2.2.1 and update PyYAML and urllib3.


sagepe:
This is deployed to staging. Before we merge, we'll need to be ready to deploy to a new server and rebuild the search index for the CAPE site, which is currently untested.

struan (approved):
makes sense, build and test fine and works locally

PR #795, opened 2026-02-09
[Scorecards] Added fix for filter where the url has a type of council

lucascumsille:
Currently, in a single section page where we display a message advising users the Combined Authorities don't have this section, the link "Show all Combined Authority sections" takes them to `/sections/?type=combined` however this URL does not activate the filter in the sections page to display only the CA.

This commit takes in account the arguments in the URL, so the right option is chosen when loading the page

https://github.com/user-attachments/assets/b71c5aab-d244-4999-baef-2d35ff002923

### DIFF
--- a/scoring/static/scoring/js/main.js
+++ b/scoring/static/scoring/js/main.js
@@ -432,6 +432,20 @@ forEachElement('[data-methodology-switch-council-type]', function(trigger){
     });
 });
 
+// Trigger council type filter from URL parameter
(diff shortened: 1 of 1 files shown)


zarino (commented):
This works, but it would be nice to do this without JavaScript somehow. Especially since we’re using a query param (which Django can access, so therefore could modify the returned page itself, no JavaScript required) 🤔

PR #825, opened 2026-08-11
[CAPS] Added newsletter to footer and about page

lucascumsille:
Fixes: https://github.com/mysociety/Design-Internal-tasks/issues/60

### About page
Screenshot 2026-08-12 at 08 17 42
Screenshot 2026-08-12 at 08 21 32

Screenshot 2026-08-12 at 08 17 49
Screenshot 2026-08-12 at 08 21 47

### Footer Desktop
Screenshot 2026-08-11 at 14 55 13

### Footer Mobile
Screenshot 2026-08-11 at 14 54 26


zarino (changes requested):
Footer looks ok to me!

I’m not thrilled about the form being stuffed into the navigation sidebar on the About page, though. (I worry it distracts from the navigation links, and looks too much like a banner advert that people will be trained to ignore.) I wonder, instead, whether we could work the newsletter signup form _into_ the “mySociety” section of that page (beside, or below, the paragraphs of descriptive text?) … and if we do that, it seems churlish not to _also_ do the same for CE UK’s newsletter too, in their section (you can get the markup for this from the Scorecards site).

I realise the CAPE About page is a low traffic page on a low traffic site, so probably not worth spending ages on this. But if it’s a relatively quick change, it’d help maintain some of the visual hierarchy on that page _and_ respect one of our partners, both at the same time.

@lucascumsille what do you think?

lucascumsille:
@zarino Sounds good, I amended the screenshots. So we include CEUK newsletter and mySociety one, in their respective sections.

zarino (approved):

PR #826, opened 2026-09-03
update to Django 5.2

struan:
Note the database update instructions in the first commit


zarino (commented):
This all worked!

Although, I note that Django told me there had been changes to the models, and when I made migrations, it altered the `council.related_authorities`, `historicalplandocument.document_type`, and `plandocument.document_type` fields. (I guess caused by some change in how Django 5.2 handles `ManyToManyField`s and `PositiveSmallIntegerField`s?)

Should we also commit that migration as part of this branch?

Or was my migration caused by my local database being out of date in some other, unrelated way?

PR #827, opened 2026-09-09
Tidy up `/script` directory

zarino:
Fixing some inconsistencies I spotted while testing out #826, taking inspiration from how we handle `script`s in the Local Intelligence Hub codebase:

- Standardise on `script/manage` rather than `python3 manage.py` and `poetry run manage.py`. Note that `POETRY_VIRTUALENVS_CREATE=false` in the Dockerfile means poetry installs into the system site-packages, so there’s no need to `poetry run` anything, `python` or `python3` works fine on its own.
- Remove old virtualenv-related stuff, we use poetry now.
- Make `script/makemigrations` run on its own.


struan (approved):
FWIW the `update_venv` script was used when deploying the site but I've updated the deploy stuff to just call poetry directly. Probably worth making sure this is deployed to the scorecards staging site to make sure it doesn't break anything with deployment

